PR #150, opened 2015-01-06
Implement OCSP support in Vulcand.

### LINKED ISSUE #134: OCSP stapling support
I would love to see [OCSP stapling](http://en.wikipedia.org/wiki/OCSP_stapling) support to reduce OCSP validation costs.

klizhentas:
## Purpose

Implement certificate revocation checking in Vulcand

http://en.wikipedia.org/wiki/OCSP_stapling

## Implementation

This PR introduces new package `stapler` that implements in-memory cache and retriever with push notifications. Mux implementation now uses Stapler and adds them to the server handshake 

Fixes #134


stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+// Clock is an optional argument to the New function, by default the system clock is used
+func Clock(clock timetools.TimeProvider) StaplerOption {
+	return func(s *stapler) error {
+		s.clock = clock
+		return nil
+	}
+}
+
+// New returns a new instance of in-memory Staple resolver and cache
+func New(opts ...StaplerOption) (Stapler, error) {
```
jeremyschlatter: The error return value is always nil. I would prefer to remove it for now. Maybe someday someone will add a StaplerOption that returns an error, but I don't expect it to be hard to update the clients if that happens.

proxy/mux.go
```diff
@@ -45,13 +46,20 @@ type mux struct {
...
+	stapleUpdatesC chan *stapler.StapleUpdated
+	stopStapleC    chan struct{}
 }
 
 func (m *mux) String() string {
 	return fmt.Sprintf("mux(%d, %v)", m.id, m.state)
 }
 
-func New(id int, o Options) (*mux, error) {
+func New(id int, st stapler.Stapler, o Options) (*mux, error) {
```
jeremyschlatter: How about making the stapler optional? You could create a default stapler with `stapler.New` if none is provided.

engine/model.go
```diff
@@ -85,9 +85,40 @@ func (a *Address) Equals(o Address) bool {
 	return a.Network == o.Network && a.Address == o.Address
 }
 
+// Sets up OCSP stapling
+type OCSPSettings struct {
+	Enabled            bool
+	Period             string
+	Responders         []string // optional responder
+	SkipSignatureCheck bool
```
jeremyschlatter: Is this here for unit tests?

engine/model.go
```diff
@@ -85,9 +85,40 @@ func (a *Address) Equals(o Address) bool {
 	return a.Network == o.Network && a.Address == o.Address
 }
 
+// Sets up OCSP stapling
+type OCSPSettings struct {
+	Enabled            bool
+	Period             string
+	Responders         []string // optional responder
```
jeremyschlatter: It's not obvious to me what "responder" means here. Can you add a short comment to clarify what this is?

engine/model.go
```diff
@@ -85,9 +85,40 @@ func (a *Address) Equals(o Address) bool {
...
+
+func (o *OCSPSettings) RefreshPeriod() (time.Duration, error) {
+	if o.Period == "" {
+		return time.Hour, nil
+	}
+	return time.ParseDuration(o.Period)
+}
+
+func (o *OCSPSettings) Equals(other *OCSPSettings) bool {
+	if o.Period != other.Period {
```
jeremyschlatter: What about o.Period == "1m", other.Period == "60s"? If you want that to compare as equal you could use `RefreshPeriod()` instead.

jeremyschlatter: Alternatively, you could make `Period` be a `time.Duration` instead of a `string`.

engine/model.go
```diff
@@ -85,9 +85,40 @@ func (a *Address) Equals(o Address) bool {
...
+		return false
+	}
+	if (len(o.Responders) == 0 || len(other.Responders) == 0) || len(o.Responders) != len(other.Responders) {
```
jeremyschlatter: I'm surprised two `OCSPSettings` structs are considered unequal if they both have 0 responders. Why is that?

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
+// package stapler implements OCSP stapling feature described here: http://en.wikipedia.org/wiki/OCSP_stapling
```
russjones: You might also want to reference the two RFC that deal with OCSP:

* https://tools.ietf.org/html/rfc2560
* http://tools.ietf.org/html/rfc6066

proxy/mux.go
```diff
@@ -74,6 +86,21 @@ func New(id int, o Options) (*mux, error) {
...
 	}
+
+	// Subscribe to staple responses and kick staple updates
+	m.stapler.Subscribe(m.stapleUpdatesC, m.stopStapleC)
+
+	go func() {
+		for {
+			select {
+			case <-m.stopStapleC:
+				log.Infof("%v stop listening for staple updates", m)
```
jeremyschlatter: stop -> stopped

seems a bit more clear to me

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+// stapler provides implementation that caches the staple on the first request and periodically updates the stapler
+package stapler
+
+import (
+	"bytes"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"github.com/mailgun/vulcand/Godeps/_workspace/src/golang.org/x/crypto/ocsp"
```
russjones: Minor style issue, can this commit be moved down so it's grouped with the rest of the Godeps.

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+	if s.clock == nil {
+		s.clock = &timetools.RealTime{}
+	}
+	go s.fanOut()
+	return s, nil
+}
+
+type stapler struct {
+	v       map[string]*hostStapler
+	mtx     *sync.Mutex
```
russjones: If you make this field anonymous then (change it to `sync.Mutex`) you can just do:

```
s.Lock()
defer s.Unlock()
```

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+		}
+	}
+	if s.clock == nil {
+		s.clock = &timetools.RealTime{}
+	}
+	go s.fanOut()
+	return s, nil
+}
+
+type stapler struct {
```
russjones: Can you add some comments to these fields. While it won't be part of the documentation, it makes working on the internals easier.

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+type Stapler interface {
+	// HasHost returns true if Stapler holds the response in cache
+	HasHost(host engine.HostKey) bool
+	// StapleHost returns the relevant StapleResponse, or error in case if response is unavailable
+	StapleHost(host *engine.Host) (*StapleResponse, error)
+	// DeleteHost deletes any OCSP data associated with the host entry
+	DeleteHost(host engine.HostKey)
+	// Subscribe subscribes the channel to the series of OCSP updates
+	Subscribe(chan *StapleUpdated, chan struct{})
+	Close()
```
jeremyschlatter:
```go
// Close closes the Stapler. The returned error is always nil.
Close() error
```

would make `Stapler` implement `io.Closer`. Might be useful someday.

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+func New(opts ...StaplerOption) (Stapler, error) {
+	s := &stapler{
+		v:           make(map[string]*hostStapler),
+		mtx:         &sync.Mutex{},
+		eventsC:     make(chan *stapleFetched),
+		closeC:      make(chan struct{}),
+		subscribers: make(map[int32]chan *StapleUpdated),
+		kickC:       make(chan bool),
+		discardC:    nil,
+		// setup transport more aggressive timeouts for OCSP staple responses
```
jeremyschlatter: Why have more aggressive timeouts?

engine/model.go
klizhentas: Nope, that's a part of the contract for ocsp

engine/model.go
klizhentas: that's a terminology used in OCSP, will add a comment

engine/model.go
klizhentas: This would make the UX kinda bad as time.Duration is translated to nanoseconds in JSON

engine/model.go
klizhentas: that's a bug

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+func (s *stapler) Subscribe(in chan *StapleUpdated, closeC chan struct{}) {
+	myId := s.subscribe(in)
+	go func() {
+		select {
+		case <-closeC:
+			s.unsubscribe(myId)
+		}
+	}()
```
jeremyschlatter: Cool. You don't need the select, though:

```go
go func() {
        <-closeC
        s.unsubscribe(myID)
}()
```

stapler/stapler.go
```diff
@@ -0,0 +1,529 @@
...
+	if !ok {
+		return
+	}
+	hs.stop()
+	delete(s.v, hk.Name)
+	log.Infof("%s deleted %v", s, hs)
+}
+
+func (s *stapler) Subscribe(in chan *StapleUpdated, closeC chan struct{}) {
+	myId := s.subscribe(in)
```
jeremyschlatter: myId -> myID

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

proxy/mux.go
klizhentas: yeah, but it's actually always used by the server, so if OCSP is turned on and the stapler is nil  that would be a configuration error, what I wanted to avoid by making it non-optional

proxy/mux.go
klizhentas: makes sense

stapler/stapler.go
klizhentas: it's just 30 seconds stdlib timeouts are too much for retrieving staple response, Otherwise sever configuration and startup would be affected

stapler/stapler.go
klizhentas: will do

stapler/stapler.go
klizhentas: Yep, but honestly I prefer explicit to anonimous fields as I'm always confused myself when missing the field using `godef` for example

testutils/testutils.go
```diff
@@ -138,6 +154,17 @@ func NewTestKeyPair() *engine.KeyPair {
 	}
 }
 
+func NewOCSPResponder() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		bytes, err := hex.DecodeString(OCSPResponseHex)
+		if err != nil {
+			panic(err)
+		}
```
jeremyschlatter: You already decoded this in init(). Why not reuse `OCSPResponseBytes` here?

klizhentas:
@russjones @jeremyschlatter addressed code review comments, lgty?

Issue #129, opened 2014-11-25
Support for self-signed certificates on upstream endpoints


technolo-g:
Due to the dynamic nature of the backends, it would be nice if one were able to use self-signed certificates for upstream endpoints. One way that I would envision this working could be specifying a CA certificate to use such as curl's --cacert option [1] or perhaps just a way to accept untrusted certs. 

[1] http://curl.haxx.se/docs/sslcerts.html

dademurphy:
+1. This would also facilitate <a href="https://github.com/bradfitz/http2">HTTP/2</a> or SPDY to backends.

klizhentas:
Here's something that works:

http://www.vulcanproxy.com/proxy.html#tls-options
http://www.vulcanproxy.com/proxy.html#https-backends

klizhentas:
Released in `v0.8.0-beta.1`

technolo-g:
Awesome, thanks @klizhentas! I'll give it a spin :)

PR #152, opened 2015-01-10
Expose TLS settings for backends and HTTPS listeners

klizhentas:
Purpose
-----------
Provide more configuration options for TLS, such as cipher suites, supported versions, session tickets and skipping certificate checks.

Implementation
--------------------
This commit exposes TLS settings for backends and HTTPS listeners via API, etcd and CLI interfaces.

Fixes
-------

### DIFF
--- a/engine/json.go
+++ b/engine/json.go
@@ -54,18 +54,6 @@ type RawMiddleware struct {
 	Middleware json.RawMessage
 }
 
-type rawListener struct {
-	Id       string
-	Protocol string
-	Address  Address
-	Scope    string
-	Settings json.RawMessage
-}
-
-type rawListenerSettings struct {
-	TLS json.RawMessage
-}
-
 func HostsFromJSON(in []byte) ([]Host, error) {
 	var hs rawHosts
 	err := json.Unmarshal(in, &hs)
@@ -115,7 +103,7 @@ func HostFromJSON(in []byte, name ...string) (*Host, error) {
 }
 
 func ListenerFromJSON(in []byte, id ...string) (*Listener, error) {
-	var rl *rawListener
+	var rl *Listener
 	err := json.Unmarshal(in, &rl)
 	if err != nil {
 		return nil, err
@@ -123,13 +111,12 @@ func ListenerFromJSON(in []byte, id ...string) (*Listener, error) {
 	if len(id) != 0 {
 		rl.Id = id[0]
 	}
-	var settings interface{}
-	if rl.Protocol == HTTPS && len(rl.Settings) != 0 {
-		if settings, err = HTTPSListenerSettingsFromJSON(rl.Settings); err != nil {
+	if rl.Protocol == HTTPS && rl.Settings != nil {
+		if _, err = NewTLSConfig(&rl.Settings.TLS); err != nil {
 			return nil, err
 		}
 	}
-	return NewListener(rl.Id, rl.Protocol, rl.Address.Network, rl.Address.Address, rl.Scope, settings)
+	return NewListener(rl.Id, rl.Protocol, rl.Address.Network, rl.Address.Address, rl.Scope, rl.Settings)
 }
 
 func ListenersFromJSON(in []byte) ([]Listener, error) {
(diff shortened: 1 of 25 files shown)


klizhentas:
Partially fixes #129

klizhentas:
@jeremyschlatter @r0mant please review

engine/json.go
```diff
@@ -103,15 +115,21 @@ func HostFromJSON(in []byte, name ...string) (*Host, error) {
 }
 
 func ListenerFromJSON(in []byte, id ...string) (*Listener, error) {
-	var l *Listener
-	err := json.Unmarshal(in, &l)
+	var rl *rawListener
```
jeremyschlatter: Why not continue using `Listener`? It has the same fields.

engine/json.go
```diff
@@ -54,6 +54,18 @@ type RawMiddleware struct {
 	Middleware json.RawMessage
 }
 
+type rawListener struct {
+	Id       string
+	Protocol string
+	Address  Address
+	Scope    string
+	Settings json.RawMessage
```
jeremyschlatter: If you embed `HTTPSListenerSettings` directly here you could remove `HTTPSListenerSettingsFromJSON` and `TLSSettingsFromJSON` below.

https://gist.github.com/jeremyschlatter/64cddb7597c8a85059ae

engine/json.go
```diff
@@ -54,6 +54,18 @@ type RawMiddleware struct {
 	Middleware json.RawMessage
 }
 
+type rawListener struct {
+	Id       string
```
jeremyschlatter: Id -> ID

engine/model_test.go
```diff
@@ -175,6 +215,43 @@ func (s *BackendSuite) TestOCSPSettingsEq(c *C) {
...
+			a: Listener{Settings: &HTTPSListenerSettings{TLS: TLSSettings{SessionTicketsDisabled: false}}},
+			b: Listener{Settings: &HTTPSListenerSettings{TLS: TLSSettings{}}},
+			e: true,
+			c: "different https",
+		},
+	}
+	for _, o := range options {
+		c.Assert((&o.a).SettingsEquals(&o.b), Equals, o.e, Commentf("TC: %v", o.c))
+	}
+}
```
jeremyschlatter: I would add some tests that expect `SettingsEquals` to return false. Otherwise

```go
func (l *Listener) SettingsEquals(o *Listener) bool {
        return true
}
```
would pass the test.

engine/model_test.go
```diff
@@ -294,3 +371,316 @@ func (s *BackendSuite) TestServerFromJSON(c *C) {
...
+
+					tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
+					tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
+
+					tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+					tls.TLS_RSA_WITH_AES_128_CBC_SHA,
+				},
+
+				InsecureSkipVerify: false,
+			},
```
jeremyschlatter: I'm surprised `C` here doesn't have something that corresponds to the SessionCache field in `S`.

engine/model_test.go
```diff
@@ -294,3 +371,316 @@ func (s *BackendSuite) TestServerFromJSON(c *C) {
...
+					tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+
+					tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
+					tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+
+					tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
+					tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
+
+					tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+					tls.TLS_RSA_WITH_AES_128_CBC_SHA,
```
jeremyschlatter: Can you keep the cipher suite spacing the same between `C` and `S`? I think that would make it easier to tell at a glance that they are the same.

engine/tls.go
```diff
@@ -0,0 +1,234 @@
...
+	// MaxVersion is max supported TLS version, "VersionTLS12" is default
+	MaxVersion string
+
+	// SessionTicketsDisabled disables session ticket resumption support
+	SessionTicketsDisabled bool
+
+	// SessionCache specifies TLS session cache, default it 'LRU' with capacity 1024
+	SessionCache TLSSessionCache
+
+	// Preferred CipherSuites, default is
```
jeremyschlatter: Unfinished comment  here.

engine/tls.go
```diff
@@ -0,0 +1,234 @@
...
+	}, nil
+}
+
+// NewTLSSessionCache validates parameters and creates a new TLS session cache
+func NewTLSSessionCache(s *TLSSessionCache) (tls.ClientSessionCache, error) {
+	cacheType := s.Type
+	if cacheType == "" {
+		cacheType = LRUCacheType
+	}
+	if cacheType != "LRU" {
```
jeremyschlatter: Can you refer to LRU the same way here and two lines above? Either `LRUCacheType` or `"LRU"`. I somewhat prefer `"LRU"`, but either way is fine.

proxy/backend.go
```diff
@@ -69,6 +69,7 @@ func (b *backend) updateSettings(be engine.Backend) error {
 		return err
 	}
 	t := newTransport(s)
+
```
jeremyschlatter: Just noting this line in case you didn't mean to insert it.

engine/model_test.go
klizhentas: I'll add a different test for it

jeremyschlatter:
lgtm

klizhentas:
@r0mant do you have any comments?

r0mant:
lgtm

PR #153, opened 2015-01-13
Fix redirect http to https

### LINKED ISSUE #120: [Enhancement]  Rewrite Plugin should rewrite http to https
Hello,
It would be cool, if vulcand could have a native way to redirect http to https.
Currently the rewrite plugin only allows to redirect a location/path to another.
It would be cool to have a Matcher that allows to match the complete URL and redirect to another like http://foo.bar to https://foo.bar or like http://foo.bar/admin/* to https://foor.bar

porjo:
Closes #120

### DIFF
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -81,7 +81,7 @@ func (rw *rewriteHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	if rw.redirect {
+	if rw.redirect && newURL != oldURL {
 		(&redirectHandler{u: parsedURL}).ServeHTTP(w, req)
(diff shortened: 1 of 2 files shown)


klizhentas:
Thanks! Can you add a test for it as well?

porjo:
> Thanks! Can you add a test for it as well?

The test [TestRewriteScheme](https://github.com/mailgun/vulcand/blob/master/plugin/rewrite/rewrite_test.go#L236) was breaking after my change, however I'm not sure what it actually achieves, so I've commented it out for now.

klizhentas:
I'll take a closer look tomorrow, thanks!

klizhentas:
I've updated the test and merged it, thanks for helping with digging the TLS bug!

PR #155, opened 2015-01-14
Add trace plugin

klizhentas:
Purpose
-----------

We need to be able to capture HTTPS access logs in structured format for debugging purposes.

Implementation
--------------------
Trace plugin outputs information about requests in structured format to any unix or udp socket. It uses syslog format to emit logs. The syslog prefix can be removed and JSON part can be put into ELK cluster.

### DIFF
--- a/Godeps/_workspace/src/github.com/mailgun/oxy/trace/trace.go
+++ b/Godeps/_workspace/src/github.com/mailgun/oxy/trace/trace.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/mailgun/vulcand/Godeps/_workspace/src/github.com/mailgun/oxy/utils"
@@ -91,20 +92,21 @@ func (t *Tracer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 }
 
 func (t *Tracer) newRecord(req *http.Request, pw *utils.ProxyWriter, diff time.Duration) *Record {
-	r := &Record{
+	return &Record{
 		Request: Request{
-			Method:  req.Method,
-			URL:     req.URL.String(),
-			TLS:     newTLS(req),
-			Headers: captureHeaders(req.Header, t.reqHeaders),
+			Method:    req.Method,
+			URL:       req.URL.String(),
+			TLS:       newTLS(req),
+			BodyBytes: bodyBytes(req.Header),
+			Headers:   captureHeaders(req.Header, t.reqHeaders),
 		},
 		Response: Response{
 			Code:      pw.StatusCode(),
+			BodyBytes: bodyBytes(pw.Header()),
 			Roundtrip: float64(diff) / float64(time.Millisecond),
 			Headers:   captureHeaders(pw.Header(), t.respHeaders),
 		},
 	}
-	return r
 }
 
 func newTLS(req *http.Request) *TLS {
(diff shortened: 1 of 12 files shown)


klizhentas:
@r0mant @jeremyschlatter can you please review?

This basically exposes the tracer as a plugin in vulcand

Godeps/_workspace/src/github.com/mailgun/oxy/trace/trace.go
```diff
@@ -0,0 +1,228 @@
...
+	return func(t *Tracer) error {
+		t.reqHeaders = append(t.reqHeaders, headers...)
+		return nil
+	}
+}
+
+// ResponseHeaders adds response headers to capture
+func ResponseHeaders(headers ...string) Option {
+	return func(t *Tracer) error {
+		t.respHeaders = append(t.reqHeaders, headers...)
```
r0mant: Shouldn't it be `t.respHeaders` here?

klizhentas: wow, nice catch

plugin/registry/registry.go
```diff
@@ -28,5 +29,9 @@ func GetRegistry() *plugin.Registry {
 		panic(err)
 	}
 
+	if err := r.AddSpec(trace.GetSpec()); err != nil {
+		panic(err)
+	}
+
```
jeremyschlatter: There are enough of these now that I think it's worth consolidating them a bit:

```go
func GetRegistry() *plugin.Registry {
    r := plugin.NewRegistry()

    specs := []*plugin.MiddlewareSpec{
        ratelimit.GetSpec(),
        connlimit.GetSpec(),
        rewrite.GetSpec(),
        cbreaker.GetSpec(),
        trace.GetSpec(),
    }

    for _, spec := range specs {
        if err := r.AddSpec(spec); err != nil {
            panic(err)
        }
    }

    return r
}
```

plugin/trace/trace.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// CliFlags is used to add command-line arguments to the CLI tool - vctl
+func CliFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:  "addr",
+			Usage: "Address of the output, e.g. syslog:///tmp/out.sock",
+		},
+		cli.StringSliceFlag{
```
r0mant: What do you provide as values for `reqHeader` and `respHeader` arguments?

klizhentas: header names to capture

plugin/registry/registry.go
klizhentas: ok

plugin/trace/trace.go
```diff
@@ -0,0 +1,205 @@
...
+	case "WARNING":
+		pr |= syslog.LOG_WARNING
+	case "NOTICE":
+		pr |= syslog.LOG_NOTICE
+	case "INFO":
+		pr |= syslog.LOG_INFO
+	case "DEBUG", "":
+		pr |= syslog.LOG_DEBUG
+	default:
+		return pr, fmt.Errorf("uknown severity: %v", sev)
```
r0mant: You return `pr` in case of error here, but `0`, not `f`, in the function below (line 200) for seemingly the same situation. Not sure this matters, but just wanted to point it out in case you want to make this consistent.

plugin/registry/registry.go
klizhentas: done

plugin/trace/trace.go
klizhentas: done

plugin/trace/trace.go
r0mant: Can I capture all headers? Don't say "list them all" :)

klizhentas: That's a good idea I will add * as a special case a bit later

r0mant:
lgtm

plugin/registry/registry.go
jeremyschlatter: I just noticed that this file is autogenerated (I think? I don't see the first-line comment in the autogeneration code). I still think it's worth changing, but it looks like the change should be in `vbundle/templates.go`.

klizhentas: Nope it's not autogenerated

plugin/trace/trace_test.go
```diff
@@ -0,0 +1,150 @@
...
+	c.Assert(err, IsNil)
+	conn, err := net.ListenUnixgram("unixgram", unixAddr)
+	c.Assert(err, IsNil)
+	defer conn.Close()
+
+	outC := make(chan []byte, 1000)
+	closeC := make(chan bool)
+	defer close(closeC)
+	go func() {
+		for {
```
jeremyschlatter: Why loop here? The code below only reads from `outC` once.

jeremyschlatter:
Can you update vbundle/template.go to match the new registry.go format?

plugin/trace/trace_test.go
klizhentas: More like a server pattern than a real reason

klizhentas:
@jeremyschlatter do you have any other comments? I've applied the previous ones

jeremyschlatter:
lgtm

PR #160, opened 2015-01-19
Fix trace package by vendoring latest oxy

klizhentas:


klizhentas:
Note: re-vendor manners

PR #161, opened 2015-01-21
Emit gaguges once a second instead of emitting individual entries

klizhentas:
Purpose
----------
Statsd becomes a bottleneck for highly concurrent and frequent requests, as it results in syscall and locking on every request

Implementation
-------------------
Emit gauges instead of individual metrics to statsd every second, speeds up the process significantly.

### DIFF
--- a/proxy/mux.go
+++ b/proxy/mux.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/url"
 	"sync"
+	"time"
 
 	"github.com/mailgun/vulcand/engine"
 	"github.com/mailgun/vulcand/stapler"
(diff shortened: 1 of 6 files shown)


klizhentas:
@anton-efimenko  @bradgignac can you please review?

proxy/mux.go
```diff
@@ -101,6 +106,20 @@ func New(id int, st stapler.Stapler, o Options) (*mux, error) {
 			}
 		}
 	}()
+
+	m.wg.Add(1)
+	go func() {
+		defer m.wg.Done()
```
anton-efimenko: Wasn't able to find m.wg.Wait() anywhere. Is it happening in a different file?

klizhentas: Wait is called in `func (m *mux) Stop(wait bool) {`

anton-efimenko: Found, thanks!

anton-efimenko:
lgmt

bradgignac:
:shipit:

Issue #154, opened 2015-01-13
travis CI


porjo:
Any reason why the Go unit tests couldn't be run through travis CI?

porjo:
Go test coverage via [Coverals](https://coveralls.zendesk.com/hc/en-us/articles/201342809-Go) might be nice too :smile:

PR #167, opened 2015-01-30
Add .travis.yml

porjo:
Relates to #154 

To enable Travis CI support, the following steps need to be taken:
1. Sign-in to [TravisCI](http://travis-ci.org/)
2. Activate Github Webhook against `mailgun/vulcand`
3. Merge this commit

Subsequent commits will then trigger a Travis build. More info [here](http://docs.travis-ci.com/user/getting-started/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: go
 go:
-        - 1.3
         - 1.4
 script:
         - go test -v ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: go
+go:
+        - 1.3
+        - 1.4
+script:
+        - go test -v ./...
+
+sudo: false
--- a/README.md
+++ b/README.md
@@ -27,5 +27,5 @@ Project info
 | blog          | http://blog.vulcanproxy.com                                     |
 | discussions   | https://groups.google.com/d/forum/vulcan-proxy                  |
 | roadmap       | [roadmap.md](ROADMAP.md)                  |
-
+| build status  | [![Build Status](https://travis-ci.org/mailgun/vulcand.svg?branch=master)](https://travis-ci.org/mailgun/vulcand) |
 


thedrow:
Can we use the docker based containers so that the build would boot faster? We can do that if we don't require sudo access.
It's a matter of specifying ```sudo: false```

porjo:
Thanks, wasn't aware of that option. I've updated `.travis.yml` now

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+go:
+        - 1.3
```
klizhentas: Vulcand does not support 1.3 actually, I suggest sticking with 1.4 for now, can you please make the change?

porjo: Oh, OK - I didn't realise that. In fact, I've been compiling and running vulcand with 1.3! What is the reason for not supporting 1.3?

porjo:
Hmm, looks like the Docker container environment is causing one of the tests to fail with:
```
/dev/log: no such file or directory
```

I must confess that I didn't test my fork using the Docker container infrastructure, but only with the default Travis environment. I expect that removing 'sudo: true' from .travis.yml will resolve that.

porjo:
I've sent pull request #168 now to fix this.

thedrow:
@porjo Can you please report this to travis?

porjo:
@thedrow I'm not sure this is a Travis issue. It's not a bug as such, as it's just a case of Docker not creating that device by default. The options I can see for resolving this inside a Docker container:

- bind mount Docker host's `/dev/log`
- permit container to run as 'root' (in order to install syslog ourselves)

I'm pretty sure the Travis team are not going to approve either of these.

Other options:

- modify the unit tests to not depend on the existence of `/dev/log`
- continue to use the standard Travis environment which has `/dev/log`

thedrow:
I don't think that unit tests should depend on the existance of any I/O device.
I think this should be mocked.
The problem will still occur with functional or integration tests.

PR #176, opened 2015-02-23
Fix circuit breaker empty response, fixes #170

### LINKED ISSUE #170: No response body from circuit breaker middleware
Hey,

I've been testing out the circuit breaker using the latest version from master at the time of writing and while I can correctly trigger the breaker and see it setting the content type and response code the body is never set for the fallback response.

I've tried using the example in the docs and this also but neither worked:

````
etcdctl set /vulcand/frontends/f1/middlewares/cb1 '{
    "Id": "cb1",
    "Priority": 1,
    "Type": "cbreaker",
    "Middleware": {
        "Condition": "NetworkErrorRatio() < 1.0",
        "Fallback": {
            "Type": "response",
            "Action": {
                "StatusCode": 400,
                "Body": "[1, 2, 3]",
                "ContentType": "application/json"
            }
        },
        "FallbackDuration": 10000000000000,
        "RecoveryDuration": 100000000,
        "CheckPeriod": 100000000
    }
}'
````

The rest of the config we have is:

````
etcdctl set /vulcand/backends/b1/backend '{"Type": "http", "Settings": {"KeepAlive": {"MaxIdleConnsPerHost": 1024, "Period": "5s"}}}'
etcdctl set /vulcand/backends/b1/servers/srv1 '{"URL": "http://httpbin.org"}'
etcdctl set /vulcand/frontends/f1/frontend '{"Id":"f1","Route":"Path(\"\/get\")","Type":"http","BackendId":"b1"}
````

I've been testing this using curl and I can see using the `-I` flag that the content type/response code are set but without that I don't get any response body.

klizhentas:
Vendor in new oxy and scroll

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -58,43 +58,43 @@
 		},
 		{
 			"ImportPath": "github.com/mailgun/oxy/cbreaker",
-			"Rev": "a888f10a61efaab5162a0fdc66fc725844862289"
+			"Rev": "563343dd5c1f1997846c85e353fbf81708982ab5"
 		},
 		{
 			"ImportPath": "github.com/mailgun/oxy/connlimit",
-			"Rev": "a888f10a61efaab5162a0fdc66fc725844862289"
+			"Rev": "563343dd5c1f1997846c85e353fbf81708982ab5"
 		},
 		{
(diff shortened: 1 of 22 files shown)


klizhentas:
@r0mant can you take a look at the PR? It's very simple

r0mant:
lgtm

PR #181, opened 2015-03-06
Case insensitive header matching fix.

Zapadlo:
- strings.ToLower() incoming host headers because we do this on the request, otherwise it won't match


klizhentas:
@Zapadlo thanks for digging into this! 

I'd suggest to move fix to github.com/mailgun/route and then I can re-vendor the new version, would you mind doing that?

Can you also add test case for this in maillgun/route? That'd be much appreciated!

Thanks,
Sasha

porjo:
I've sent a pull request to mailgun/route now using @Zapadlo's fix

PR #187, opened 2015-03-14
Fix Content-Type for JSON

### LINKED ISSUE #179: Invalid Content-Type "application-json" in HTTP response
See [this line](https://github.com/mailgun/vulcand/blob/66d7adb6aee0b8de7150e71f7fe7d0b02e6aae47/proxy/mux.go#L607). The string should instead be [something like this](http://stackoverflow.com/questions/477816/what-is-the-correct-json-content-type).

porjo:
Closes #179


klizhentas:
@porjo looks good, can you add a test too?

porjo:
Test added

klizhentas:
Thanks!

PR #200, opened 2015-04-07
Declare phony targets in Makefile

glkz:
All the targets in makefile seems to be phony targets. So, I added a ```.PHONY``` section.


jessespears:
That's correct, thanks for contributing!

PR #202, opened 2015-04-08
Allow adding a custom NotFound middleware to Vulcan.

archisgore:
WHAT:
    1. Inside mux.go is a NotFound http.Handler that is used
       as the catch-all handler for requests that do not
       match any route.
    2. Repeatedly, we face situations where we want to know
       such requests and act on them (i.e. either throw a
       friendlier 404 Not Found page, or redirect them,
       or audit/log them to find missing/erraneous routes
       in our system.
    3. This is distinctly different from using the CircuitBreaker
       that only operates on routes that exist, but fail.

HOW:
    1. mux.go already has a NotFound handler, which I renamed to
       DefaultNotFound (for clarity should anyone read the code in
       the future and try to make sense of where the default
       one resides.)
    2. I then modified proxy.Options to hold a Middleware that could
       be set by the caller (service.go who instantiates mux router)
    3. service.go itself obtains this middleware from Registry which
       also holds all MiddlewareSpecs.
    4. I added methods to the registry to add/get the default middleware.

Wrote some tests to go along with the changes where necessary.

WHY NOT:
    1. Why not use etcd/engine? - Two reasons: Major one being time.
       I need this change to use vulcand ASAP. However, the major
       reason was because I need a NotFound handler when routes are
       failing - i.e. waiting for etcd to SET a middleware would
       be a point of failure. Given that I'm already compiling vulcan
       using vbundle, this seemed like a reliable way to ensure I have
       something in there, if all else fails.

    2. Why not pass a MiddlewareSpec, instead of Middleware? -
       Mainly because instantiating middlewares requires some
       parameters - which for other middlewares are obtained from
       etcd/engine. However, this middleware must be instantiated
       and ready to go with as little ceremony as possible, in order
       for it to be the catch-all default handler.

FUTURE IMPROVEMENTS:
    1. I may add an etcd or engine-based path as well. But we'll
       need to discuss a new key/namespace where we can add
       "default" or "catch-all" middlewares and write
       logic around injecting them.

### DIFF
--- a/plugin/middleware.go
+++ b/plugin/middleware.go
@@ -53,14 +53,13 @@ type SpecGetter func(string) *MiddlewareSpec
 
 // Registry contains currently registered middlewares and used to support pluggable middlewares across all modules of the vulcand
 type Registry struct {
-	specs              []*MiddlewareSpec
-	notFoundMiddleware Middleware
+	specs    []*MiddlewareSpec
+	notFound Middleware
 }
 
 func NewRegistry() *Registry {
 	return &Registry{
-		specs:              []*MiddlewareSpec{},
-		notFoundMiddleware: nil,
+		specs: []*MiddlewareSpec{},
 	}
 }
 
@@ -91,13 +90,13 @@ func (r *Registry) GetSpecs() []*MiddlewareSpec {
 	return r.specs
 }
 
-func (r *Registry) AddNotFoundMiddleware(notFoundMiddleware Middleware) error {
-	r.notFoundMiddleware = notFoundMiddleware
+func (r *Registry) AddNotFoundMiddleware(notFound Middleware) error {
+	r.notFound = notFound
 	return nil
 }
 
 func (r *Registry) GetNotFoundMiddleware() Middleware {
-	return r.notFoundMiddleware
+	return r.notFound
 }
 
 func verifySignature(fn interface{}) error {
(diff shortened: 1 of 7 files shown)


archisgore:
Is there something I can do to help this move along? I'm using this in production for quite a time, and it appears to be stable so far.

I'm happy to change/modify/discuss any concerns you may raise.

plugin/middleware.go
```diff
@@ -53,12 +53,14 @@ type SpecGetter func(string) *MiddlewareSpec
 
 // Registry contains currently registered middlewares and used to support pluggable middlewares across all modules of the vulcand
 type Registry struct {
-	specs []*MiddlewareSpec
+	specs              []*MiddlewareSpec
+	notFoundMiddleware Middleware
```
klizhentas: Can you rename it to notFound for simplicity as the Middleware is already enforced by the type

plugin/middleware.go
```diff
@@ -53,12 +53,14 @@ type SpecGetter func(string) *MiddlewareSpec
...
 }
 
 func NewRegistry() *Registry {
 	return &Registry{
-		specs: []*MiddlewareSpec{},
+		specs:              []*MiddlewareSpec{},
+		notFoundMiddleware: nil,
```
klizhentas: I think explicit initialization is redundant in this case, as `nil` is implied by default:

http://play.golang.org/p/erfsuZ1-Y8

klizhentas:
Aside from a couple of minor comments, looks good to me

PR #219, opened 2015-06-24
/tmp fillup bugfix

martingartonft:
grab fix from :
https://github.com/mjgarton/multibuf/commit/b9a1261a65e5c6c6f835bcee2e16df4b47323b32


tboerger:
This should be fixed upstream first and than we need to update the Godeps reference.

martingartonft:
I did submit this upstream first, but there has been no response.

tboerger:
@klizhentas what's up with the upstream project?

klizhentas:
turns out I forgot to set `watching` in github, sorry just merged it

archisgore:
I'm going to push this change as a re-vendor of the mailgun/multibuf package
https://github.com/vulcand/vulcand/commit/f0953ab9e0b8d6bc7189ac3bdb8485ebe5d19793

PR #220, opened 2015-06-26
Remove backend from backends array, when it goes out of ETCD

archisgore:
During our high-frequency change scenario (we use vulcand programmatically),
we discovered a weird edge condition on backends still living in the backends
array and thus causing old servers to be hit.

The crux of the problem was this:
1. PUT /backends/b1/servers/s1
2. PUT /frontends/f1/frontend {backendId: b1}
3. DELETE /frontends/f1
4. DELETE /backends/b1
5. PUT /backends/b1/servers/s2
6. PUT /frontnends/f1/frontned {backendId: b1}

Now if we hit frontend f1, we mysteriously saw hits to server s1.

Upon deeper investigation I found that in the mux.go:DeleteBackend function,
the backend was never really REMOVED from the m.backends map.

This meant that when upsertBackend is called, m.backends[bk] retrieves
the old backend, which is "updated" with the new backend, instead
of a completely new backend being created.

This left over some servers in the backend (either becuase etcd failed
to publish events or something went wrong.) Either way, if a frontend DID
hold a reference, from an outside observer, merging of a new etcd backend
with an old one makes little sense. AKA

1. PUT /backends/b1/servers/s1
2. PUT /frontends/f1/frontend {backendId: b1}
3. DELETE /backends/b1
4. PUT /backends/b1/servers/s2
5. PUT /frontneds/f2/frontend {backendId: b1}

While I totally admit this scenario is illegal, I DO believe
that frontend f2 should get the backend that exists in etcd at the time
the frontend was created. It cannot be expected to have memory of what
other users may or may not have done in the past. Basically, if I had
a worker process, it should be able to delete everything
and reset it from scratch to get the state it needs.

By explicitly removing that backend from the m.backends map, it means
if a frontend f2 is added without FIRST having added a backend, a proper
error will be raised, and if a backend did exist, THAT is the latest
backend it would hold a reference to.

A note about the DeferClose:
1. It is a simple form of reference-counting. When the backend is deleted,
we tell it that it no longer is referenced.
2. When the last frontend that references it is unlinked, it will simply
close its idle connections.

### DIFF
--- a/engine/memng/mem.go
+++ b/engine/memng/mem.go
@@ -13,10 +13,11 @@ import (
 
 // Mem is exported to provide easy access to its internals
 type Mem struct {
-	Hosts     map[engine.HostKey]engine.Host
-	Frontends map[engine.FrontendKey]engine.Frontend
-	Backends  map[engine.BackendKey]engine.Backend
-	Listeners map[engine.ListenerKey]engine.Listener
+	Hosts             map[engine.HostKey]engine.Host
+	Frontends         map[engine.FrontendKey]engine.Frontend
+	Backends          map[engine.BackendKey]engine.Backend
+	Listeners         map[engine.ListenerKey]engine.Listener
+	SharedMiddlewares []engine.Middleware
 
 	Middlewares map[engine.FrontendKey][]engine.Middleware
 	Servers     map[engine.BackendKey][]engine.Server
@@ -179,6 +180,7 @@ func (m *Mem) UpsertMiddleware(fk engine.FrontendKey, md engine.Middleware, d ti
 	if _, ok := m.Frontends[fk]; !ok {
 		return &engine.NotFoundError{Message: fmt.Sprintf("'%v' not found", fk)}
 	}
+
 	defer func() {
 		m.emit(&engine.MiddlewareUpserted{FrontendKey: fk, Middleware: md})
 	}()
--- a/engine/memng/mem.go
+++ b/engine/memng/mem.go
@@ -13,11 +13,10 @@ import (
 
 // Mem is exported to provide easy access to its internals
 type Mem struct {
-	Hosts             map[engine.HostKey]engine.Host
-	Frontends         map[engine.FrontendKey]engine.Frontend
-	Backends          map[engine.BackendKey]engine.Backend
-	Listeners         map[engine.ListenerKey]engine.Listener
-	SharedMiddlewares []engine.Middleware
+	Hosts     map[engine.HostKey]engine.Host
+	Frontends map[engine.FrontendKey]engine.Frontend
+	Backends  map[engine.BackendKey]engine.Backend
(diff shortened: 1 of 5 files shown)


engine/memng/mem.go
```diff
@@ -13,10 +13,11 @@ import (
...
 type Mem struct {
-	Hosts     map[engine.HostKey]engine.Host
-	Frontends map[engine.FrontendKey]engine.Frontend
-	Backends  map[engine.BackendKey]engine.Backend
-	Listeners map[engine.ListenerKey]engine.Listener
+	Hosts             map[engine.HostKey]engine.Host
+	Frontends         map[engine.FrontendKey]engine.Frontend
+	Backends          map[engine.BackendKey]engine.Backend
+	Listeners         map[engine.ListenerKey]engine.Listener
+	SharedMiddlewares []engine.Middleware
```
klizhentas: looks like this property is not needed for this change

klizhentas:
the fix looks good to me, and thanks for the detailed explanation. I will spend some more time thinking about `DeferClose` before merging it in though.

The situation when the backend is being deleted while being referenced should lead to this piece of code to be executed:

https://github.com/mailgun/vulcand/blob/master/proxy/mux.go#L412

But in absense of transactions, I can see a race that can happen, when user will call:

```
* Delete frontend
* Delete backend
```

But we may see:

```
Delete backend (fails)
Delete frontend (succeeds)
````

This will lead to backend that stays forever and is not being deleted, but not to the scenario that you mentioned though. Can you elaborate a bit more on the case when `DeferClose` can occur? 

Have you seen it in practice?

engine/memng/mem.go
archisgore: Ah good catch. You of course know where this came from. :-) It's a bad merge from my changes to do shared middlewares.

archisgore: I'll remove it.

archisgore:
Perhaps the DeferClose isn't needed. when I read that function, I assumed that this line MUST execute before a backend is let lose: https://github.com/mailgun/vulcand/blob/master/proxy/mux.go#L416

And when a fronted still references the backend, the line you posted, returns from the function. I wasn't sure if NOT calling http.transport.CloseIdleConnections() would end up with a memory leak, so I added a deferClose, which tells the backend it is no longer referenced. So it uses the number of frontends as a reference-counter, and when the counter hits zero, it calls b.Close() on itself.

If the backend is sure to be cleaned up correctly, then we don't need the DeferClose thingy. Let me know.

archisgore:
Also, what's with the broken build? I'm glad to help out to fix it in any way I can.

archisgore:
There you go - the change is a one-liner and very simple. If you could pull it in soon, we can repoint our code to the mailgun/vulcand repo.

klizhentas:
thanks!

PR #224, opened 2015-07-14
Added X-Forward-Proto support in rewrite reles

paulbakker:
We use Vulcan behind a AWS Elastic Load Balancer for SSL offloading. For HTTP traffic we want to redirect to HTTPS, but because SSL is already handled by the ELB, the wrong URL is constructed in rewrite.go.

This fix uses the X-Forwarded-Proto header that is used by most proxies.

### DIFF
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -156,14 +156,14 @@ func CliFlags() []cli.Flag {
 
 func rawURL(request *http.Request) string {
 	scheme := "http"
-	if request.TLS != nil || isXForwardedHttps(request) {
+	if request.TLS != nil || isXForwardedHTTPS(request) {
 		scheme = "https"
 	}
 
 	return strings.Join([]string{scheme, "://", request.Host, request.RequestURI}, "")
 }
 
-func isXForwardedHttps(request *http.Request) bool {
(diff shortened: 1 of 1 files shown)


plugin/rewrite/rewrite.go
```diff
@@ -156,12 +156,19 @@ func CliFlags() []cli.Flag {
 
 func rawURL(request *http.Request) string {
 	scheme := "http"
-	if request.TLS != nil {
+	if request.TLS != nil || isXForwardedHttps(request) {
```
klizhentas: should be `isXForwardedHTTPS`

plugin/rewrite/rewrite.go
```diff
@@ -156,12 +156,19 @@ func CliFlags() []cli.Flag {
...
 		scheme = "https"
 	}
+
 	return strings.Join([]string{scheme, "://", request.Host, request.RequestURI}, "")
 }
 
+func isXForwardedHttps(request *http.Request) bool {
```
klizhentas: `isXForwardedHttps` -> `isXForwardedHTTPS`

klizhentas:
overall looks good, just a minor rename and I'll merge

archisgore:
Looks like the original author renamed the method. Merging this in.

PR #227, opened 2015-08-06
Use CoreOS's upstream etcd instead of forked copy

pquerna:
Also updates mailgun/scroll (open PR: https://github.com/mailgun/scroll/pull/32 )


tboerger:
There have been patches on top of the forked etcd, maybe this stuff is still not fixed upstream?

pquerna:
The diff is here:

https://github.com/coreos/go-etcd/compare/master...mailgun:master

@klizhentas has this open PR for coreos/go-etcd that contain the commits:

https://github.com/coreos/go-etcd/pull/121

pquerna:
So, Vulcand doesn't even use `AddChildDir`.... so the diff is kinda pointless.

PR #232, opened 2015-08-18
Export SeverityFlag, so others can override it in the options struct.

pquerna:

PR #233, opened 2015-08-18
InitWithConfig needs an explicit severity

wrotki:
Otherwise logger creation fails on severity parsing
And hence there are no logs whatsoever


klizhentas:
Thanks a lot, this if fixed by  #232 as I believe

wrotki:
Cool, thanks.

PR #235, opened 2015-08-27
Re-vendor oxy to include PassHostHeader changes

archisgore:
These changes are an outcome of this issue: https://github.com/mailgun/vulcand/issues/172


klizhentas:
thanks!

tboerger:
Does that fix the forwarded headers for the hostname?

PR #237, opened 2015-08-28
fix rewrites for encoded urls

patrickbcullen:
This patch fixes two bugs with the rewrite plugin

1. The rewrite plugin was executing even if the URL did not match the Regexp parameter of the rewrite plugin config. It most cases this had not impact because the code did not cause a rewrite. However, if there was a bug in the logic it would cause a rewrite of the URL which was unexpected. To fix this I added a MatchString call to make sure the URL matches before doing any processing.

2. The RequestURI should be the encoded value of the URL Path. If the Path had spaces in it (i.e. localhost/foo%20bar) the code was unencoding this and then updating the RequestURI with an invalid URI with a space in it. The fix was using the builtin url.RequestURI() function which returns the encoded version of the URL. Th url.RequestUIR function also addes the RawQuery parameters to the Request URI so I deleted that extra code from this function.

### DIFF
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -65,6 +65,12 @@ func newRewriteHandler(next http.Handler, spec *Rewrite) (*rewriteHandler, error
 func (rw *rewriteHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	oldURL := rawURL(req)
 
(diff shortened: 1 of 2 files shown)


klizhentas:
@r0mant can you take a look at this just to make sure it won't break any Mailgun's stuff?

r0mant:
lgtm

PR #239, opened 2015-09-04
Added ability to sync cluster with Etcd periodically.

archisgore:
The purpose of this change is to keep vulcand resilient against
long-running etcd clusters. This change allows vulcan to periodically
synchronize the latest members of the etcd cluster from the
member it is current connected to, this giving it fallback
members to use in case the primary member were to fail for
whatever reason.

This allows us to maintain the etcd cluster by adding new replacement machines,
and taking down older machines, but so long as the cluster itself
is running overall, vulcand instances talking to it, will continue to work.

### DIFF
--- a/plugin/rewrite/rewrite.go
+++ b/plugin/rewrite/rewrite.go
@@ -65,6 +65,12 @@ func newRewriteHandler(next http.Handler, spec *Rewrite) (*rewriteHandler, error
 func (rw *rewriteHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	oldURL := rawURL(req)
 
+	// only continue if the Regexp param matches the URL
+	if !rw.regexp.MatchString(oldURL) {
+		rw.next.ServeHTTP(w, req)
+		return
+	}
+
 	// apply a rewrite regexp to the URL
 	newURL := rw.regexp.ReplaceAllString(oldURL, rw.replacement)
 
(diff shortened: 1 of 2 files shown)


tboerger:
:+1:

archisgore:
Ok hold on. I've broken some test. I can repro it locally. Wait for a fix. Working on it ASAP.

mischief:
@archisgore there is a caveat to be aware of - if you point etcd client at a etcd proxy and call SyncCluster, your client url list will change to the upstream etcd servers, and not the proxy.

mischief:
@archisgore also fyi, vulcand should probably switch to `github.com/coreos/etcd/client` at the earliest opportunity.

archisgore:
Yup. That's precisely what I expect to happen.

philips:
I would suggest that you sync cluster on some reasonable interval like 60s by default instead of disabling.

tboerger:
And i think it should be disabled because of backwards compat.

philips:
@tboerger fair enough

archisgore:
Ok I figured out the cause of this UT failure. It's a race condition where, in some cases,

supervisor_test.go: TearDownTest() calls Supervisor.Stop() which closes s.restartC channel.

However, at around the same time, the engine.Subscribe() goroutine is sending a nil to the restartC channel which has already been closed.

klizhentas:
@archisgore thanks for finding this race condidtion! can you add a separate issue about it, so we can address it?

PR #242, opened 2015-09-11
This is a reissue of a previous PR with a UT race condition fix.

archisgore:
This adds the optional capability to SyncCluster with etcd (described in detail in  the commit message.)

The bigger change which took me some time to figure out was the race condition causing UTs to fail occasionally (described in detail in the other commit message.)

I believe with the UT fix, two of the other pending PRs should pass as well.

### DIFF
--- a/engine/etcdng/etcd.go
+++ b/engine/etcdng/etcd.go
@@ -17,41 +17,47 @@ import (
 )
 
 type ng struct {
-	nodes    []string
-	registry *plugin.Registry
-	etcdKey  string
-	client   *etcd.Client
-	cancelC  chan bool
-	stopC    chan bool
-	logsev   log.Severity
-
-	options Options
+	nodes            []string
+	registry         *plugin.Registry
+	etcdKey          string
+	client           *etcd.Client
+	cancelC          chan bool
+	stopC            chan bool
+	syncClusterStopC chan bool
+	logsev           log.Severity
+	options          Options
 }
 
 type Options struct {
-	EtcdConsistency string
-	EtcdCaFile      string
-	EtcdCertFile    string
-	EtcdKeyFile     string
-	Box             *secret.Box
+	EtcdConsistency         string
+	EtcdCaFile              string
+	EtcdCertFile            string
+	EtcdKeyFile             string
+	EtcdSyncIntervalSeconds int64
+	Box                     *secret.Box
 }
 
 func New(nodes []string, etcdKey string, registry *plugin.Registry, options Options) (engine.Engine, error) {
 	n := &ng{
-		nodes:    nodes,
-		registry: registry,
-		etcdKey:  etcdKey,
-		cancelC:  make(chan bool, 1),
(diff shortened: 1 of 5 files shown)


engine/etcdng/etcd.go
```diff
@@ -729,6 +733,13 @@ func (n *ng) deleteKey(key string) error {
 	return convertErr(err)
 }
 
+func (n *ng) periodicallySyncCluster() {
+	for {
```
klizhentas: this goroutine should exit when `Close` has been called. I suggest adding close channel and doing

```
select {
     case time.After <-...
      case <- n.closeC <-...
```

archisgore:
Thanks for the feedback Alexander. Does this look good to you?

klizhentas:
spent some time about `restartC` panic on race condition. The fact that we need to guard it with the mutex seems a bit wrong - after all, channels are sync mechanism. How about instead, we will make buffered channel for restartC (e.g. 100)- and do not close it at all? in this case, when stop will be called, we can make sure that no panic will occur and the goroutine would not block too.

archisgore:
That was my original idea too. When I don't close the channel restartC, the UT hangs. You can try it for yourself if you check out head to my commit, and then remove the part where I close restartC.

I don't know the code well enough to know who/what is waiting on it. I know the sole listener on that channel does break successfully though. I'm happy to investigate a bit later.

klizhentas:
oh, it's been a while since I wrote this chunk, you are right, here's what is waiting:

https://github.com/mailgun/vulcand/blob/master/supervisor/supervisor.go#L240

klizhentas:
ok, in this case we can do the following:

* intro new channel - `broadcastCloseC` - it will be used to signal all interested parties that we are about to shutdown. We will only close it in `Stop`
* `supervise` will be changed to select for both `broadcastCloseC` and `restartC`:

```go
select {
   case <- restartC: // do real restart
   case <- broadcastCloseC: // call stop()
}
```
this will let `supervise` to exit without blocking on `restartC`. What do you think?

archisgore:
Yup. Sounds simple enough. I'll have this done tonight or early tomorrow. I hate locking too. I love this approach. Thanks for the suggestion!

klizhentas:
@archisgore thanks for spending your time on this problem!

archisgore:
I think this should do it. I introduced the same situation by adding delays and it works every time consistently. No more locks!

archisgore:
Pinging for reminder again. This change too I've been running in production for over two weeks and works very reliably.

klizhentas:
@archisgore sorry for the delay. One last comment - let's make `broadcastC` a buffered channel to avoid the case when multiple goroutines created in `init` can be potentially blocked on the channel forever once `supervise` returns

archisgore:
You got it! Coming up in 10 minutes.

archisgore:
It's done. All checks passed!

klizhentas:
lgtm, thanks for your efforts!

PR #245, opened 2015-09-16
Ability to disable Default listener

archisgore:
This is a backwards compatible change that allows you to disable the default listener from being created, by forcing a boolean parameter to false.


archisgore:
Pinging again for a reminder. I'm using this change on production for quite a while now and works very reliably and consistently.

klizhentas:
lgtm

PR #247, opened 2015-09-24
First iteration of making Router pluggable.

archisgore:
I filed an ISSUE earlier to discuss this here: https://github.com/mailgun/vulcand/issues/218

The crux of this change is:
1. To make Router an interface behind which implementations can be
changed out.

2. To allow the Registry to specify an implementation.

3. To fall back to the default implementation (route.Mux)
when one is not provided by the registry.

This change also depends on this pending PR in the mailgun/route library
https://github.com/mailgun/route/pull/10

WHY?

As we try to add new features to mailgun/route, we continue to have to work off custom branches
and re-vendor the files, and rebuild vulcand for each experiment. For example we have
a forked route library which contains massive amounts of logging.

Having re-vendored files is always confusing because we don't know what branch/fork it came
from. We'd much rather have a way to host 10 forks and simply plug in an implementation
at runtime based on command-line flags.

This allows us to rapidly innovate, and I'm sure gives vulcand users the ability to
use richer or simpler languages.

For instance, one experiment I want to run is use of Javascript-based expressions
which would allow the use of VERY rich matching rules at the cost of
VERY high compute load per rule-match. It's just an idea, but trying it out
can be made easier by making it pluggable.

As it happens this change doesn't affect how Vulcan works today, and the defaults
do not change.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -102,7 +102,7 @@
 		},
 		{
 			"ImportPath": "github.com/mailgun/route",
-			"Rev": "85cf368efd247dabbcc2e53272ef4276f94a87c4"
+			"Rev": "20d2530fb5fba644d31762ad8f0581643ee75368"
 		},
 		{
 			"ImportPath": "github.com/mailgun/scroll",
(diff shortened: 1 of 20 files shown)


archisgore:
As an FYI here, we've used these changes for over two weeks in production by now. Works perfectly.

PR #251, opened 2015-10-13
Websocket

gamorejon:
Add support for websocket frontends and backends using a modified version of this: https://gist.github.com/sheenobu/5aad1b1c07e9ff52eefe


tboerger:
So I need to define a separate frontend for the websocket connection?

gamorejon:
Yeah and a separate backend. Both with "ws" as the type.

tboerger:
A separate backend as well? That sucks pretty hard... No registrator adds multiple servers for a single docker container, that's for sure a use case for multiple people. Maybe you should look at traefik, you need to upgrade the connection only. I think a separate type is somehow overkill.

gamorejon:
Sorry, it's not actually a strict requirement for the backend. For the frontend the requirement is pretty artificial and I can remove it pretty easily. My understanding was that vulcan prefers that you define a frontend per route anyhow and I figured websockets would probably involve different middleware stacks. Websockets won't work with much of the current stuff as it's making heavy use oxy which is making heavy use of things like RoundTripper which doesn't support websockets. I'll take it out though since it's not really doing anything.

tboerger:
But anyway you can compare it with traefik as it's using oxy as well :)

tboerger:
For the records: https://github.com/EmileVauge/traefik/pull/62

gamorejon:
Thanks, traefik looks clean. It looks like we are doing basically the same thing only he/she is probably doing a better job at copying the http headers (will add that when I get some spare cycles). Both of us are basically hijacking the connection before it hits the oxy. It seems to me, the correct long-term solution though, is to actually make oxy at least websocket aware.

archisgore:
I'll begin reviewing this code sometime late this month. Have you tested/run this in a production scenario yet?

gamorejon:
Yes, but for the very specific use-case of dynamically matching client/server ssh-over-websocket connections.

pcross616:
any update on this?  Love to see this get merged.

klizhentas:
we've merged websocket support into oxy, the underlying load balancing library, the next step would be to revendor it in vulcand and we should have websockets working

DefunctExodus:
@klizhentas  Great to hear, Really need to be able to use web sockets

archisgore:
I don't think this is relevant anymore. The oxy websockets support is vendored into vulcan. I'm closing this issue. Please reopen of it's still relevant.

PR #254, opened 2015-10-16
Remove godeps, rewrite import paths to use direct dependencies

pquerna:


tboerger:
As I already said I really dislike this kind of unco trollable dependencies.

pquerna:
@tboerger I guess the major pivot point on this is *how* people are using vulcand  -- we are using it as a library (via vbundle) so we can tweak behavoirs at compile time. Is this the majority use case, or are many people directly using vulcand?

tboerger:
All people I know are using vulcand as a program and not as a lib. Even as a lib the new vendoring works fine

klizhentas:
@pquerna I actually think it's a valid point. I plan to move to 1.5, use VENDOREXPERIMENT and remove explicit dependency on godeps

klizhentas:
closing if favor of #278

PR #263, opened 2015-11-23
Migration to new etcd client

archisgore:
This change does the following:
1. Forks the old etcdng into the new etcdv2ng (Version 2 of etcd engine).
   This is intentional to allow a one-line change to move everyone back
   to the tried and tested, and known-to-work reliable etcdng which has NO
   code changes made to it.

2. Pulled in the latest github.com/coreos/etcd/client package and Godepped
   it into this package. That will allow taking dependencies on it to make future changes.

3. Re-vendored uroji/go/codec, coreos/go-etcd/client to ensure all package versions compile.

This ensures that should there be a problem, switching to etcdng would be simple and clean, even if there
were commits made ahead of this change, possibly making it irrevertible.


archisgore:
We tested this and the new client works more efficiently in watching notifications too. Merging it in now.

Issue #267, opened 2015-11-29
docker-build fails to build working image on OS X 


shrinathaithal:
`make docker-build` on OS X statically compiles the app, but for OS X. This fails to run docker container, throws error similar to this - 
`Error response from daemon: Cannot start container c0e4ba53f6919f4f2a31ad85e3848b64afbed7e79064bf5e0fcfbb44b824c1b4: [8] System error: exec format error
make: *** [docker-run-fast] Error 1`

PR #268, opened 2015-11-29
Fix docker image building from scratch on OS X

shrinathaithal:
When making minimal docker image on OS X, the docker container fails to start. This fixes it by compiling for linux.


shrinathaithal:
Fixes #267

archisgore:
That's a good catch! Thanks. I'm merging it in.

PR #273, opened 2015-12-17
When creating a new proxy, call Stop() on it if anything goes wrong

### LINKED ISSUE #216: Debug and resolve memory leak
How can we debug and resolve some memory leak within vulcand? Even a vulcand instance without any traffic consumes multiple GB of memory after just a few days runtime.

pquerna:
If your etcd cluster is having a bad day, `Supervisor` will try to restart everything, and if `initProxy` has an error talking to etcd (or any other error), which it most definitely can, you will leak a partially initialized proxy/mux instance -- this is because in `proxy/mux.go`, on `New`, two goroutines are created, and they will keep the reference to the `mux` around until `Close` is called.

Don't love the style used here with a defer function, but this seemed the easiest way to get it done.

Should fix or at least partially fix #216

### DIFF
--- a/supervisor/supervisor.go
+++ b/supervisor/supervisor.go
@@ -144,10 +144,18 @@ func (s *Supervisor) TopServers(key *engine.BackendKey) ([]engine.Server, error)
 
 func (s *Supervisor) init() error {
 	proxy, err := s.newProxy(s.lastId)
+	s.lastId += 1
 	if err != nil {
 		return err
 	}
-	s.lastId += 1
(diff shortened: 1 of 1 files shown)


pquerna:
We have been running this in our environment for 10 hours, seems to have resolved the memory leak discussed in #216

archisgore:
Makes sense to me. Also looks like a good thing to do regardless.

klizhentas:
lgtm

PR #277, opened 2016-01-29
Use Sirupsen/logrus instead of mailgun/log

agalitsyn:
Hi!

https://github.com/Sirupsen/logrus is nice package which can bring feature-rich logging in vulcand.
In this PR I've tried to replace mailgun/log without breaking backward compability and prepare repo for adding some logging configuration.

Pros from this PR:
4 logger are available: console, syslog, json, logstash
And more severity levels: panic, fatal, error, warning, info, debug


agalitsyn:
CI is confusing me. It says:

`FAIL: supervisor_test.go:92: SupervisorSuite.TestGracefulShutdown
supervisor_test.go:107:
    c.Assert(GETResponse(c, b.FrontendURL("/")), Equals, "Hi, I'm endpoint")
supervisor_test.go:166:
    c.Assert(err, IsNil)
... value *url.Error = &url.Error{Op:"Get", URL:"http://localhost:11800/", Err:(*net.OpError)(0xc20805d100)} ("Get http://localhost:11800/: dial tcp 127.0.0.1:11800: connection refused")`

But in other tests with this endpoint tests passed. Not reproducable at local  dev env

archisgore:
This one I'll leave to original vulcan devs and mailgun team.

However, I would like this change a LOT! I personally use/prefer logrus and have many hooks/extensions I'd love to take advantage of.

chanaoka:
+1. already using logrus elsewhere. love it and would be great to have the same capabilities from vulcan.

klizhentas:
@agalitsyn thanks! I will review this weekend and ping back!

agalitsyn:
Thanks guys! To be honest I'm starting golang programming recently (I'm a python developer) and I'm really want to hear some feedback on code.

There are 4 issues that bothers me:
1. Libraries that still use mailgun/log. The biggest is oxy library.
2. Failing CI, I can't get why it don't pass. On my work and home env all tests passed, I can't find obvious mistakes.
3. I've tried to create log package inside a vulcan, which can hide logger implementation details and provide unified struct. So as for me it's better than direct logrus imports, because we can do any changes in log package without changing all files =) But I came up with some blocker - 
If I create and return struct Logger which points to logrus.Logger, I should use import "github.com/vulcand/vulcand/log", and init object like log.Logger.New(some settings here). And initialize it each time i need logger. And use like log.Logger.Infof()
If I want to leave code like log.Infof(), which most files contatins, I should reimplement lazy object creation and interface like in [logrus/exported.go(https://github.com/Sirupsen/logrus/blob/master/exported.go) 

Any thoughts on there issues? We can plan it as task and implement in independent PR

klizhentas:
@agalitsyn feel fee to send PR for OXY as well.

agalitsyn:
Created https://github.com/vulcand/oxy/pull/29

PR #278, opened 2016-01-31
switch to go15 and GO vendor experiment

klizhentas:
This commit switches Vulcand to use Go 1.5 as a default build environment and uses Go vendor experiment supported only by versions of Go 1.5 and older


klizhentas:
@pquerna please review and merge if it looks good

agalitsyn:
LGTM, when you plan to add this to master?

PR #279, opened 2016-02-03
Use GO15VENDOREXPERIMENT with godeps

agalitsyn:
Hello

SInce vulcand default golang version is 1.5 and [godep supports](https://github.com/tools/godep#go-15-vendor-experiment) GO15VENDOREXPERIMENT I suggest to move all deps to vendor/

Also golang 1.6 is public beta already.
> Go 1.6 keeps the vendoring support, no longer considered experimental, and enables it by default.

Ref https://tip.golang.org/doc/go1.6


klizhentas:
rejected in favor of #279

PR #280, opened 2016-02-08
Fix broken dependency on codegangsta/cli

agalitsyn:
Hi

Since project moved to `GO15VENDOREXPERIMENT=1` after https://github.com/vulcand/vulcand/pull/279 dependency `github.com/codegangsta/cli` is broken for middlewares

It happens because we have tree
```
- foo.go
  - vendor/
    - a/
    - b/
        - vendor/a/
```

So we should remove `github.com/codegangsta/cli` from vendoring


agalitsyn:
No, wrong take. Removing this dependency actually breaks tests and build, because nobody fetch it now.

So I close PR and create issue to discuss https://github.com/vulcand/vulcand/issues/281

PR #282, opened 2016-02-25
protect from '401: The event in requested index is outdated and cleared'

iamruinous:
We have a very busy etcd server which causes many 401s to be sent to vulcand. When vulcand supervisor reloads, it causes random 504s from our upstream proxy (AWS ELB in this case).

This commit ignores the 401 and re-creates a Watcher.

### DIFF
--- a/engine/etcdv2ng/etcd.go
+++ b/engine/etcdv2ng/etcd.go
@@ -13,13 +13,14 @@ import (
 	"time"
 
 	"crypto/x509"
+	"io/ioutil"
+
 	log "github.com/Sirupsen/logrus"
 	etcd "github.com/coreos/etcd/client"
 	"github.com/vulcand/vulcand/engine"
 	"github.com/vulcand/vulcand/plugin"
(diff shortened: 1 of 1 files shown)


klizhentas:
hey, sorry for the delay in review. I will dig into this

horkhe:
We just recently run into the same issue. We have about 300 frontends and about 1000 backend servers registered. But ignoring the error from Etcd is probably not the best solution. That would result in inconsistency between the distributed system state and Vulcand's view on it. Which in turn may result in all sorts of hard to debug issues.

We solved it by speeding up update processing in vulcand with #326 and #328. And reducing the number of updates to Etcd. In particular we made our services hearbeat via [TTL refresh API](https://coreos.com/etcd/docs/latest/v2/api.html#refreshing-key-ttl) that does not trigger watches (requires Etcd v2.3.7 or higher). if you are using Etcd for something else then you may consider deploying an Etcd cluster dedicated for vulcand.

Issue #261, opened 2015-11-16
Add Support for TLS Client Authentication


felipegs:
Like https://github.com/vulcand/vulcand/issues/250 It will be good if we can configure it, because configure it on middleware it not works on protocol level it is application level, so it is less secure.

PR #285, opened 2016-04-19
Added client certificate support for backend

upcFrost:
Added additional client certificate tls options for backends. Can be used for example when the receiving side (upstream) performs certificate-based authentication.

Not sure if it was possible to make it on the middleware level since middleware works with HTTP requests while SSL connection is established on the lower level.

Also, it addresses the https://github.com/vulcand/vulcand/issues/261

**Short description and limitations**: This code assumes that certificates are stored **in the same place** across the installation and have the **same name**. In my case this assumption holds, but it would fail in the general case. 
I'm not sure how to make the same trick for the general case, probably by adding the ClientCert field to the host node.

### DIFF
--- a/engine/tls.go
+++ b/engine/tls.go
@@ -111,18 +111,23 @@ func NewTLSConfig(s *TLSSettings) (*tls.Config, error) {
 
 	// Load CA certificates
 	caCertPool := x509.NewCertPool()
-	for _, certfile := range s.CACertFile {
-		caCert, err := ioutil.ReadFile(certfile)
-		if err != nil {
-			return nil, fmt.Errorf("Failed to open CA cert file %s: %s", certfile, err)
+	if len(s.CACertFile) != 0 {
+		for _, certfile := range s.CACertFile {
+			caCert, err := ioutil.ReadFile(certfile)
+			if err != nil {
+				return nil, fmt.Errorf("Failed to open CA cert file %s: %s", certfile, err)
+			}
+			caCertPool.AppendCertsFromPEM(caCert)
 		}
-		caCertPool.AppendCertsFromPEM(caCert)
 	}
 
 	// Load client certificate
-	cert, err := tls.LoadX509KeyPair(s.ClientCertFile, s.ClientKeyFile)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to load client key/cert pair: %s", err)
+	cert := tls.Certificate
+	if s.ClientCertFile != "" && s.ClientKeyFile != "" {
+		cert, err = tls.LoadX509KeyPair(s.ClientCertFile, s.ClientKeyFile)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to load client key/cert pair: %s", err)
+		}
 	}
 
 	return &tls.Config{
--- a/engine/tls.go
+++ b/engine/tls.go
@@ -122,7 +122,7 @@ func NewTLSConfig(s *TLSSettings) (*tls.Config, error) {
 	}
 
 	// Load client certificate
-	cert := tls.Certificate
+	var cert tls.Certificate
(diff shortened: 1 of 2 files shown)


engine/tls.go
```diff
@@ -102,6 +109,22 @@ func NewTLSConfig(s *TLSSettings) (*tls.Config, error) {
 		}
 	}
 
+	// Load CA certificates
+	caCertPool := x509.NewCertPool()
+	for _, certfile := range s.CACertFile {
+		caCert, err := ioutil.ReadFile(certfile)
+		if err != nil {
+			fmt.Errorf("Failed to open CA cert file %s: %s", certfile, err)
```
klizhentas: return here or continue to avoid adding nil file

engine/tls.go
```diff
@@ -102,6 +109,22 @@ func NewTLSConfig(s *TLSSettings) (*tls.Config, error) {
...
+		}
+		caCertPool.AppendCertsFromPEM(caCert)
+	}
+
+	// Load client certificate
+	cert, err := tls.LoadX509KeyPair(s.ClientCertFile, s.ClientKeyFile)
+	if err != nil {
+		fmt.Errorf("Failed to load client key/cert pair: %s", err)
```
klizhentas: missing return?

upcFrost: True, sry, fixed

ottodashadow:
Is there a reason this request is still open? Would backend client certificate support be accepted if it used a sealed KeyPair stored in etcd the way the host certificates work?

upcFrost:
>Is there a reason this request is still open? 

It's up to maintainer ;) also, #261 is still open.

>Would backend client certificate support be accepted if it used a sealed KeyPair stored in etcd the way the host certificates work?

Imo this way is good, but it also requires additional encryption step. Also, each backend will still require some kind of structure to separate services/endpoints/machines.

Sorry if I'm mixing things up, I already forgot about this pr:[

ottodashadow:
> sealing the backend client keypair "also requires additional encryption step"

true, but the hosts already employ a Sealed KeyPair, so the backend would just need to duplicate the idea.

> Also, each backend will still require some kind of structure to separate services/endpoints/machines.

I don't quite understand why, seeing as we can have multiple frontends, tied to multiple backends/servers. So, if vulcand needs to use a different client certificate for `/api/path1` and `/api/path2` just make two frontends with two backends.

My understanding of the vulcand components and how they fit in client certificate.

- **hosts**: hold the KeyPair that a listener uses, and gets unsealed from etcd in the [GetHost function](https://github.com/vulcand/vulcand/blob/60da89453ed1387bd8f3590bd8d3c27be87bc0cc/engine/etcdv2ng/etcd.go#L189)
- **listeners**: this is where we would store a ClientCAs pem cert/bundle, and a ClientAuth: tls.RequireAndVerifyClientCert value to solve #261 
- **frontends**: not relevant to client auth.
- **middleware**: if the listener was configured with `&tls.Config { ClientAuth: tls.RequireAndVerifyClientCert }` then we could access the list of peer certificates through `http.Request.TLS.PeerCertificates` and use that to allow or deny a request.
- **backends**: this is where we could store the client certificate that connects to the backend service.
- **servers**: not relevant to client auth. If we can assume that while a single backend may have multiple servers that vulcand load balances to, those servers are identical and as such allow the same client certificates.

I've done end to end client certificate auth in my own fork you can take a look at the [listener commit](https://github.com/ottodashadow/vulcand/commit/e5d238796beb7afb20de1147eaa42912204b544e) and the a [backend commit](https://github.com/ottodashadow/vulcand/commit/5fa463deb7c49a889057fdb1236e7c281ae243cf).

thrawn01:
Closed due to age, If someone is interested in this feature make a new PR or update this one and mention me in the PR. I'll make sure it gets merged.

Issue #75, opened 2014-09-13
Allow configuration by environment variables


robszumski:
It would be great to be able to configure vulcand with environment variables that have similar names to the flags:

`-etcd` or `$VULCAND_ETCD`
`-port` or `$VULCAND_PORT`

dcarley:
Another use case for this is preventing the sensitive value of `-sealKey` from appearing in the process list for non-root users.

PR #286, opened 2016-04-21
Added env config support

upcFrost:
Adopted coreos etcd envconfig lib to be used in vulcand.
Mixed config is possible, flags have priority to the env vars
Example usage:
`VULCAND_APIINTERFACE=0.0.0.0 VULCAND_PORT=8185 vulcand -logSeverity debug`

Note: ParseCommandLine() renamed to ParseRunOptions() for the sake of correctness since it parses not only CLI input, but also env vars now

Addressed in https://github.com/vulcand/vulcand/issues/75

### DIFF
--- a/service/options.go
+++ b/service/options.go
@@ -4,11 +4,15 @@ import (
 	"flag"
 	"fmt"
 	log "github.com/Sirupsen/logrus"
+	"github.com/coreos/etcd/pkg/flags"
+	"os"
 	"strings"
 	"time"
 )
 
 type Options struct {
+	*flag.FlagSet
+
 	ApiPort      int
 	ApiInterface string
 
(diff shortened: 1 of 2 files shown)


upcFrost:
Build fails because godep pulls an old version of the flags pkg where 'ETCD_' prefix was hardcoded. Changed in https://github.com/coreos/etcd/commit/a120ca16c03eab305b2ce5318e2249119abce442

service/options.go
```diff
@@ -4,11 +4,15 @@ import (
...
 	"fmt"
 	log "github.com/Sirupsen/logrus"
+	"github.com/coreos/etcd/pkg/flags"
+	"os"
 	"strings"
 	"time"
 )
 
 type Options struct {
+	*flag.FlagSet
```
horkhe: It does not need to be embedded in options. `fs` should be just a local variable of `ParseRunOptions`.

service/options.go
```diff
@@ -94,45 +98,53 @@ func validateOptions(o Options) (Options, error) {
...
-	flag.StringVar(&options.StatsdAddr, "statsdAddr", "", "Statsd address in form of 'host:port'")
+	fs.StringVar(&options.StatsdPrefix, "statsdPrefix", "", "Statsd prefix will be appended to the metrics emitted by this instance")
+	fs.StringVar(&options.StatsdAddr, "statsdAddr", "", "Statsd address in form of 'host:port'")
 
-	flag.BoolVar(&options.DefaultListener, "default-listener", true, "Enables the default listener on startup (Default value: true)")
+	fs.BoolVar(&options.DefaultListener, "default-listener", true, "Enables the default listener on startup (Default value: true)")
 
-	flag.Parse()
+	options.FlagSet.Parse(os.Args[1:])
+	err = flags.SetFlagsFromEnv("VULCAND", fs)
```
horkhe: I would make it the other way around - command line arguments should override environment variables.

horkhe:
And `github.com/coreos/etcd/pkg/flags` needs to be vendored.

thrawn01:
Closed due to age, If someone is interested in this feature make a new PR or update this one and mention me in the PR. I'll make sure it gets merged.

PR #292, opened 2016-06-07
Support chunking/streaming

archisgore:
This is the initial revision that supports this issue: https://github.com/vulcand/oxy/pull/45

Under "very specific use cases" (defined below), this implementation isn't beautiful and smooth. But for most common sense use cases such as large file downloads, MPEG streams, etc. etc. this solution should work just fine.

The change is simple:
1. In the FrontendSettings structure, I added a new field called "Stream". By default it is false, and defaults to the "Buffer" implementation in oxy. When set to true, it goes to a different code-path (which supports chunked encoding passthrough all the way to the client.)

The major code change is coming from the diff within Oxy, which is being reviewed here: https://github.com/vulcand/oxy/pull/45

This change allows me to test, refine and fix the streaming part of vulcan without affecting the existing hardened code path.

Now for the special use case that didn't work just beautifully?

This is the sample app:

  http.HandleFunc("/chunked", func(w http.ResponseWriter, r *http.Request) {
    flusher, ok := w.(http.Flusher)
    if !ok {
      panic("expected http.ResponseWriter to be an http.Flusher")
    }
    for i := 1; i <= 10000; i++ {
      fmt.Fprintf(w, "Chunk #%d\n", i)
      flusher.Flush() // Trigger "chunked" encoding and send a chunk...
      time.Sleep(500 * time.Millisecond)
    }
  })

Now if you connected to this app through telnet, you'd see the chunks as they are emitted and flushed.

For some reason, through vulcan, the chunks are being reformed and resized and flushed differently.

This is a very artificially simulated use-case to emulate... say a single long-lived connection sending small numbers of bytes periodically. It is a real-world usecase that I'm going to fix ASAP, but it shouldn't hold up the chunking from being merged which would allow it to be tested and hardened.


archisgore:
I'm going to merge this in. Like always, we're going to deploy this to production ASAP and begin heavy testing. Please do send ANY feedback you may have about changes you'd like and I'll fix them ASAP.

Conversely, reverting is also an option if I've broken something. I've tested this intensely. The flakey UTs are something I'm going to devote more time to next week. There's some weirdness on when a test client  might get an unexpected EOF back.

PR #299, opened 2016-09-24
Fix broken plugin/trace test

gchaincl:
plugin/trace test will fail if '/dev/log' is not present.
This file might be absent on:
* *BSD and Darwin; syslog uses '/var/run/log'
* Windows machines
* Linux using RSyslog with default options

This PR creates a unix socket to test against it


archisgore:
Thanks for making this work. Merging it in.

PR #300, opened 2016-09-24
Set a default limit for the Top API

gchaincl:
As Top Api controller is strconv.Atoi() on ?limit param, we need to specify a default value, otherwise calling the Api without a limit, the call will error returning Atoi() error.


archisgore:
LGTM.

PR #301, opened 2016-09-25
Speed up proxy initialization

horkhe:
### Problem

We noticed that sometimes it takes proxy up to 5 minutes (!!!) to startup on our system. Turned out the reason for that was inefficient initial configuration loading from Etcd, that was performed one key at a time.

### Solution

`GetSnapshot` method was introcuded to `engine.Engine` interface. That supposed to efficiently read entire configuration from an engine. This method is only implemented for `etcdv2ng` engine. For other engines stub implementations are provided returning `nil`. On proxy initialization we first call `GetSnapshot` and if `nil` is returned then resort to old style initialization logic.

This solution brought the startup time from about 5 minutes to about 30 seconds on our system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: go
 go:
-        - 1.5
-        - 1.6
-        - 1.7
+  - 1.5.4
+  - 1.6.3
+  - 1.7.1
 env:
-        - GO15VENDOREXPERIMENT=1
+  - GO15VENDOREXPERIMENT=1
 script:
-        - go test $(go list ./... | grep -v '/vendor/')
+  - go test -v $(go list ./... | grep -v '/vendor/')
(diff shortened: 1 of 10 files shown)


archisgore:
Thanks for doing this! This was on my list to do eventually, since we faced a similar problem against long-living etcd clusters with lots of config. I'll review the code later today.

archisgore:
I like the changes. I have a few things I'd love to get clarified:

1. I don't know very much about idiomatic golang. I was thinking returning "nil, nil" can get confusing for future readers of code.

Basically the expectation is that if err is nil, then the result value is always usable. If GetSnapshot isn't supported by an engine, it would be an error, that the caller would decide what to do with.

Perhaps have a shared "SnapshotNotSupportedError" that engines can return, and can be compared against?

2. What do you think about guarding against changes made *during* snapshotting by attaching the etcd change index at time of snapshot into the snapshot? Then after the snapshot processing, run start the change-watcher from that index to ensure there are no lost updates.

This will ensure that making changes while vulcan is reading config won't get missed.

horkhe:
Both comments are fair. I will implement the proposed changed.

horkhe:
@archisgore done

archisgore:
@horkhe Thanks for doing this! In addition to the startup-time improvement, this also resolves a long-existing race condition when you update keys during loading.

LGTM.

PR #302, opened 2016-09-26
Update cli.Action -> cli.ActionFunc signature

gchaincl:
As of urfave/cli version 1.15 `cli.App.Action` signature was deprecated, and instead we should use `func(*cli.Context) error`, as defined by `cli.ActionFunc`.
This commit update the funciton signature.

### DIFF
--- a/vctl/command/print.go
+++ b/vctl/command/print.go
@@ -106,3 +106,7 @@ func (cmd *Command) printFrontend(f *engine.Frontend, ms []engine.Middleware) {
 func writeS(w io.Writer, v string) {
 	w.Write([]byte(v))
 }
+
+func PrintError(cmd *Command, err error) {
+	cmd.printError(err)
+}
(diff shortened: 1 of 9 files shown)


gchaincl:
fixed by #305

horkhe:
Oops, I did not realised there was a PR for that already. Sorry about that. I should have been more observant. By the way I did not fix the unit tests, only `vctl`. So you if you do not mind please rebase your changes and submit a PR with those. I will merge it.

gchaincl:
Sure @horkhe!

vctl/command/print.go
```diff
@@ -106,3 +106,7 @@ func (cmd *Command) printFrontend(f *engine.Frontend, ms []engine.Middleware) {
 func writeS(w io.Writer, v string) {
 	w.Write([]byte(v))
 }
+
+func PrintError(cmd *Command, err error) {
+	cmd.printError(err)
```
horkhe: Just capitalize `printError` and use it in `vctl/main.go` there is no point in introducing an exported wrapper.

gchaincl: well, it makes sense

horkhe:
@gchaincl thank you!

PR #303, opened 2016-09-26
Do not log to stdout if --log=syslog

horkhe:
### Problem

It seems illogical that when vulcand is started with `--log=console` it writes logs to **stdout**, and when `--logs=syslog` it writes logs to both **stdout** and **syslog**.

### Solution

If `--log=syslog` is specified then logs are only written to **syslog**.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -71,22 +71,30 @@ func NewService(options Options, registry *plugin.Registry) *Service {
 
 func (s *Service) Start() error {
 	if s.options.Log == "console" {
+		log.SetOutput(os.Stdout)
 		log.SetFormatter(&log.TextFormatter{})
 	} else if s.options.Log == "syslog" {
-		hook, err := logrus_syslog.NewSyslogHook("", "", syslog.LOG_INFO, "")
+		devNull, err := os.OpenFile("/dev/null", os.O_WRONLY, 0)
+		if err != nil {
+			panic(err)
+		}
+		log.SetOutput(devNull)
+		hook, err := logrus_syslog.NewSyslogHook("", "", syslog.LOG_INFO|syslog.LOG_MAIL, "vulcand")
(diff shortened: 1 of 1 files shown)


gchaincl:
I think that's the intention, to log always to `stdout`, and optionally to `syslog`, but I might be wrong.

horkhe:
It is the intention introduced with logrus. Which I personally think is wrong. For it should be possible to enable/disable any logger type. In fact that is how it was when Vulcand was using mailgun/log. So this PR restores the original behaviour.

service/service.go
```diff
@@ -71,22 +71,30 @@ func NewService(options Options, registry *plugin.Registry) *Service {
 
 func (s *Service) Start() error {
 	if s.options.Log == "console" {
+		log.SetOutput(os.Stdout)
 		log.SetFormatter(&log.TextFormatter{})
 	} else if s.options.Log == "syslog" {
-		hook, err := logrus_syslog.NewSyslogHook("", "", syslog.LOG_INFO, "")
+		devNull, err := os.OpenFile("/dev/null", os.O_WRONLY, 0)
+		if err != nil {
+			panic(err)
```
klizhentas: looks good, but instead of panic can you simply fallback to stdout here? In many cases syslog socket is not exposed (e.g. when running in a container)

horkhe: It seems to me that in that case the right solution would be to update the logging parameter you start your vulcand with to remove syslog and use something else explicitly. Falling back to console is not obvious, for the user specified syslog for a reason and they will be scratching their head wondering why it is not working.

klizhentas: this argument makes sense, but in practice that means that all new containers with vulcand inside will crash, so we have no option here but to fallback to stdout to preserve backwards compatibility.

PR #306, opened 2016-09-30
Speed up GetFrontends and GetBackends functions

horkhe:
We observed that `vctl frontend ls` had taken forever to complete on our system. Turned out it was slow for the same reason as #301. I should have fixed this issue in the same PR. Well... better late than never.


horkhe:
@klizhentas @archisgore can some of you guys take a look please.

PR #307, opened 2016-10-02
Added Options.LogFormatter property

chanaoka:
Add support for custom logrus formatters. If this property is set, vulcan will use this value when calling log.SetFormatter; otherwise, the default behavior (using the Options.Log string to determine the logrus formatter) will be used.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -70,7 +70,36 @@ func NewService(options Options, registry *plugin.Registry) *Service {
 }
 
 func (s *Service) Start() error {
-	s.initLogger()
+	// if .LogFormatter is set, it'll be used in log.SetFormatter() and .Log will be ignored.
+	if s.options.LogFormatter != nil {
+		log.SetFormatter(s.options.LogFormatter)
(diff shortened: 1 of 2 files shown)


service/service.go
```diff
@@ -70,7 +70,36 @@ func NewService(options Options, registry *plugin.Registry) *Service {
...
+	} else {
+		switch s.options.Log {
+			case "console": {
+				log.SetFormatter(&log.TextFormatter{})
+			}
+			case "json": {
+				log.SetFormatter(&log.JSONFormatter{})
+			}
+			case "syslog": {
+				hook, err := logrus_syslog.NewSyslogHook("", "", syslog.LOG_INFO, "")
```
klizhentas: we should never fail if syslog is not available, check out the code in `initLogger` for detail.

archisgore:
I know @chanaoka is a little busy with other things. I'm taking this PR forward here: https://github.com/vulcand/vulcand/issues/313

Closing this PR.

PR #310, opened 2016-10-04
Run tests with race detector enabled

horkhe:
This PR enables race detector in tests. With the first run a couple of races were detected. Both are minor, but I fixed them anyway. One was actually in mailgun/manners, and another in `mux.String()`


horkhe:
@klizhentas @archisgore any comments?

PR #311, opened 2016-10-11
Expose api http client so it can be replaced

nvartolomei:
This of course would introduce backward incompatibilities for anyone using `api.Client` struct. Which could be fixed by adding a new method to client that will check if `api.Client.httpClient` is null and if it is then use the `&http.DefaultClient{}`.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -20,12 +20,13 @@ import (
 const CurrentVersion = "v2"
 
 type Client struct {
-	Addr     string
-	Registry *plugin.Registry
+	Addr       string
+	Registry   *plugin.Registry
+	HttpClient *http.Client
 }
 
 func NewClient(addr string, registry *plugin.Registry) *Client {
-	return &Client{Addr: addr, Registry: registry}
+	return &Client{Addr: addr, Registry: registry, HttpClient: http.DefaultClient}
 }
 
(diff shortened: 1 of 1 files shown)


horkhe:
What is the point of this?

horkhe:
This does not seem to be necessary. Please reopen if you can provide a reason.

nvartolomei:
@horkhe to customize http client. Even basic things like timeouts, security settings.

horkhe:
@nvartolomei that makes sense. But I do not like exposing internals to the outside world. In fact Addr and Registry are exposed for no reason. I suggest adding optional functional parameter to `New`, e.g.:

```Go
type Client struct {
    addr       string
    registry   *plugin.Registry
    httpClt *http.Client
}
  		  
type option func(c *Client) 

func WithHttpClient(httpClt *http.Client) option {
    return func(c *Client) {
        c.httpClient = httpClt
    }
}

func NewClient(addr string, registry *plugin.Registry, opts... option) *Client {
    c := Client{addr: addr, registry: registry}
    for _, opt := range opts {
        opt(c)
    }
    return &c
} 
```

So if you need to customize client you can:

```
vulcandClt := NewClient(addr, registry, WithHttpClient(httpClt))
```

Or something along these lines.

Issue #296, opened 2016-08-31
Unit tests sometimes failed with Error "EOF"


archisgore:
In some UTs under the proxy package,

NewReader(conn).ReadString("\n")

will return an error "EOF". Basically there's a premature EOF before "\n" is found.

I added plenty of debugging/logging statements to both manners and oxy/testutils Responders, but couldn't figure this out. This issue is mainly for me to go nail down what's happening.

As a side note, whenever an OCSPResponse is sent back by the server, the server apparently does not call the server-side handler. Aka, a connection either gives back an OCSP response, or the response body, but not both.

I don't know if this is a golang implementation limitation, an HTTP spec, or a bug in oxy. I'll figure this out early next week too.

PR #314, opened 2016-10-15
Vulcand no longer starts up if there is no config in etcd

epipho:
Attempting to start Vulcand for the first time with no data in etcd fails. The error log will print Service exited with error: service start failure: 100: Key not found (/vulcand) [7477]  and the application will exit.

This is due to GetSnapshot (introduced in e3f4eef8ad24f8ad46afc0af98a852f957d809f1) not checking for key not found.  

A similar regression was introduced in bc167c679135941224818ac183d9e9c634af6033 which caused GetBackends and GetFrontends to fail with a similar error.  This caused commands such as `vctl backend ls` to throw an Internal Server Error if no backends had been installed.

GetSnapshot now returns an empty snapshot if the key is not found.
GetFrontends and GetBackends now return an empty slice if the key is not found, same as before the change.


epipho:
Test failure appears to be related to #296

gchaincl:
is it possible to merge this?

archisgore:
LGTM

Issue #312, opened 2016-10-11
Support Etcd 3.x


archisgore:
I'm going to begin the initial implementation for an etcdv3ng. Fortunately our engines are swappable so this can be done safely alongside working code, and tested without affecting production/stable code paths.

https://coreos.com/blog/etcd3-a-new-etcd.html

PR #315, opened 2016-10-18
Support for Etcd ClientV3

archisgore:
This work is towards issue: https://github.com/vulcand/vulcand/issues/312

EtcdV3 is fundamentally different from V2. Read more here: https://coreos.com/blog/etcd3-a-new-etcd.html

TL;DR:
1. There are no more directories. All keys are searched by prefixes. "/foo/bar" is simply a key in and of itself. "/foo" would be an independent key.
2. Instead of directory-based TTLs to invalidate sub-keys, there are "Leases". You acquire a lease then create keys under that lease. When the lease expires all keys bound to the lease expire. This does allow for interesting cases where all frontends/backends/middlewares/servers could be under one lease (as opposed to TTLs on different directories.)

What this change does:
1. Uses the etcdv3 client to adapt a flat key structure (but syntactically the familiar vulcand key structure) into vulcand's internals.

2. All unit tests pass. The Unit tests are not modified in any fundamental way (except increase in one timeout.)

3. Added an etcd version option (Defaults to 2), so that V3 support can be opted into. Otherwise the V2 engine is used, thus allowing 100% backcompat.

4. Some Godep revendoring happened to pull in the new client.


archisgore:
I'm going to merge this in. The entire code path for v3 client is in a separate package+directory, and there's only a single branch which only kicks in when the user selects -etcdApiVer=3 flag on the command-line.

I've run this through many scenarios, and will push all fixes upstream if/when I detect any errors.

(BTW, the performance gain is significant.)

PR #323, opened 2017-04-13
Add Proxy Protocol support to Listeners

pquerna:
* Adds new option for Listeners: `ProxyProtocol`.
* Changes `HijackListener` in `mailgun/manners` https://github.com/mailgun/manners/pull/14 so that we can re-construct our listener chain (TCP Options, then HAProxy Protocol, then TLS)


klizhentas:
@pquerna merge at will

PR #324, opened 2017-04-13
Extract signal handling from service.Service

russellhaering:
Vulcand is well factored for use as a library. The main pain point I've hit is that once you've called `Start()` on a service there is no way to stop it - it directly registers a bunch of signal handlers which are used to trigger shutdown.

This change abstracts signals into `ControlCode`s, a channel of which must be passed to `Start()`. The behavior of `service.Run()` is unchanged - signal handlers are registered as before, and each signal is mapped to the corresponding control code.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -31,13 +31,47 @@ import (
 	"github.com/vulcand/vulcand/supervisor"
 )
 
+type ControlCode int
+
+const (
+	ControlCodeGracefulShutdown ControlCode = iota
+	ControlCodeImmediateShutdown
+	ControlCodeForkChild
+)
+
+func waitForSignals() chan ControlCode {
+	sigC := make(chan os.Signal, 1024)
+	signal.Notify(sigC, os.Interrupt, os.Kill, syscall.SIGTERM, syscall.SIGUSR2)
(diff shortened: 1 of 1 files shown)


klizhentas:
Good idea @russellhaering feel free to merge this in

service/service.go
```diff
@@ -31,13 +31,47 @@ import (
...
+		for {
+			signal := <-sigC
+			log.Infof("Got signal '%s'", signal)
+
+			switch signal {
+			case syscall.SIGTERM, syscall.SIGINT:
+				controlC <- ControlCodeGracefulShutdown
+			case syscall.SIGKILL:
+				controlC <- ControlCodeImmediateShutdown
+			case syscall.SIGUSR2:
```
gchaincl: why don't you handle this along with `case syscall.SIGTERM, syscall.SIGINT`

russellhaering: Good catch, this case had a typo: it was sending `ControlCodeGracefulShutdown` wghen it should have been sending `ControlCodeForkChild`. Fixed in aa74c54.

gchaincl: :+1:

PR #328, opened 2017-04-28
Lazy frontend initialization

horkhe:
### Problem
In a large system with hundreds of frontends and even more backend servers, where vulcand is installed on many hosts that share the same etcd backend, vulcand instances tend to grow very large in size. One of the reasons is that even though only a subset of frontends is used by each particular instance, all of them are initialized with pretty heavy data structures.
### Solution
The frontend data structures has to be initialized on demand when a particular frontend is used for the first time. In our case this solution helped to reduce memory consumption by 15-35% depending on how many frontends were actually used by an instance.
### Implementation details
The proxy package was heavily redesigned to achieve this goal the majority of changes were made in `mux.go`, `backend.go`, `frontend.go`. The most prominent changes are:
1. Cyclic dependency `backend<->frontend` and `mux<->frontend` were broken to `frontend->backend` and `mux->frontend` respectively. Which is much easier to grasp;
2. Finer grain mutexes were introduced in backends and frontends to ensure that lazy initialization of frontends that happens in HTTP handlers now has consistent view on the mux matrix state;
3. Lots of renamings, I just could not figure out when the code was talking about proxy/backend and when about engine/backend;

As a bonus a regression introduced in PR327 that caused vulcand deadlock on stop was fixed.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/vulcand/vulcand/stapler"
 	"github.com/vulcand/vulcand/supervisor"
 	"github.com/vulcand/vulcand/testutils"
-
 	. "gopkg.in/check.v1"
 )
 
--- a/engine/model.go
+++ b/engine/model.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/vulcand/oxy/buffer"
 	"github.com/vulcand/oxy/memmetrics"
 	"github.com/vulcand/route"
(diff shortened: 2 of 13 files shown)


horkhe:
Anybody what to take a look at this? I am going to merge it tomorrow.

archisgore:
Ive been following your commits - and missed when you removed the WIP tag. LGTM. In fact i'm waiting on your to merge it to use it. :-)

horkhe:
@archisgore we have been running it in production for more than week, so it is already battle-tested :). Although we also removed RTWatcher in our bundle to reduce memory consumption even further.

PR #329, opened 2017-04-30
Drop dependency on mailgun/scroll

horkhe:


horkhe:
@archisgore well, actually `vctl` command talks to vulcand via API. Thanks anyway.

PR #331, opened 2017-05-04
Log to syslog via UDP

horkhe:
### Problem
If vulcand is running in container with `--log=syslog`. Then it connects to syslog via Unix socket that has to be mapped to the container as a volume from the host.  But if syslog is restarted on the host that makes the mapped unix socket file invalid and vulcand starts spitting errors to console log for every attempt to log something.
### Solution
When syslog is configured as logging destination we should send messages to its UDP port 514. 
### Misc
When syslog is selected we also do not need to duplicate everything to console.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -106,40 +106,7 @@ func NewService(options Options, registry *plugin.Registry) *Service {
 }
 
 func (s *Service) Start(controlC chan ControlCode) error {
-	// if .LogFormatter is set, it'll be used in log.SetFormatter() and .Log will be ignored.
-	if s.options.LogFormatter != nil {
-		log.SetFormatter(s.options.LogFormatter)
-	} else {
-		switch s.options.Log {
-		case "console":
-			{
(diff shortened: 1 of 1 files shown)


klizhentas:
that's true, but in container it's better to log to stdout/stderr anyways

horkhe:
@klizhentas if I do that then the container starts to grow indefinitely. because the console log is every growing and never truncated. How do you deal with that?

klizhentas:
@horkhe in docker stdout and stderr are swapped and are output to the log directory where they can be rotated and forwarded by various parties (e.g. logrotate )

https://sandro-keil.de/blog/2015/03/11/logrotate-for-docker-container/

horkhe:
@klizhentas thanks, good to know. I will create a ticket for our infrastructure guys to configure that. In the meantime it won't hurt to log to syslog via UDP ;)

horkhe:
@klizhentas if you do not object I will merge this.

Issue #333, opened 2017-05-09
Dependency on mailgun/manners


horkhe:
Mailgun will no longer support its fork of https://github.com/braintree/manners. We only used it to achieve graceful HTTP server shutdown and now we have that out-of-the-box in Golang 1.8. So vulcand project has two options: either create its own fork of the repo or absorb pieces that it needs. I do not have enough knowledge in what manners features are needed by vulcand so I cannot make this call. @klizhentas @archisgore @pquerna what is going to be your opinion on that?

PR #334, opened 2017-05-09
Fixed godep reference to mailgun/manners

archisgore:
This is a cosmetic change, but allows downstream
dependencies to pull the correct versions of
packages vulcand depends on.


horkhe:
Beware of #333.

archisgore:
Hah! yes true that. I'm just getting it to a building consistent state. And I plan to try and refactor manners out - but it'll need more time. I don't fully understand what it adds. As I pull each dependency out, I'll know more.

PR #335, opened 2017-05-09
Add trustForwardHeader command line parameter

horkhe:
### Problem
Sometimes you need to run Vulcand behind another proxy. In that case `X-Forwarded-` headers should be trusted by Vulcand. Before this change, the headers could be made trusted by a per-frontend configuration knobe `TrustForwardHeader` set to `true`. But in most cases it would be easier to just enable that setting at the process level. Because usually you either run vulcand behind another proxy or not.

### Solution
Command line parameter `trustForwardHeader` was added. So that if either `trustForwardHeader` command line parameter or `TrustForwardHeader` frontend parameter is `true` then Vulcand leaves `X-Forwarded-Proto` and `X-Forwarded-Host` intact and augments `X-Forwarded-For` so that new value is constructed as `<old value>, <vulcand ip address>`.

### DIFF
--- a/engine/model.go
+++ b/engine/model.go
@@ -204,6 +204,10 @@ func (h *Host) GetId() string {
 	return h.Name
 }
 
+func (h *Host) Key() HostKey {
+	return HostKey{Name: h.Name}
+}
+
 // Frontend is connected to a backend and vulcand will use the servers from this backend.
 type Frontend struct {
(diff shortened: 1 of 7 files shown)


horkhe:
silence implies consent :)

klizhentas:
yeah, go for it

PR #336, opened 2017-05-22
Remove dependency on mailgun/manners

archisgore:
Since not ALL builds of vulcan will be on go1.8, this is a stop-gap solution where I've brought in the GracefulServer and GracefulListener code from manners, and removed the dependency on manners.

This gets us one step closer to controlling vulcan's destiny within vulcan (easier forking, less dependency worry, when we DO remove the graceful parts, etc.)

I renamed GracefulServer/GracefulListener to Server/Listener since they live in the "graceful" package.

Also moved from manner's "waitGroup" to instead use "sync.WaitGroup".

This change also piggybacks an oxy update I made which fixes WebSockets over TLS and removes a ton of redundant code. I'll keep this WIP until that code gets merged in. https://github.com/vulcand/oxy/pull/73


horkhe:
Sorry, I am swamped at work a bit. I hope to take a look a this in a day or two.

archisgore:
Don't rush it. I'm waiting for my changes in Oxy to get merged in. And I'm running this change in production, so I'll know more about stability next week.

archisgore:
This is ready for review. We tested this build in production for 2 weeks now. BTW, client-facing HTTP 2.0 works reliably now. If you expose vulcan over a TLS Listener, you'll be auto-negotiated to h2.

The sooner this gets merged in, the sooner I want to do backend http 2.0, and a full bi-directional 2.0 proxy (including server push propagation.)

archisgore:
Thanks! And good point on the tests. I will be sure to bring them in.

The change to enable H2 upgrades for clients (when they're already on TLS) was as simple as this:
https://github.com/vulcand/vulcand/pull/336/commits/5fa60bb2b6f815cd8b4d790fa1b4905dd4473679

It was recently added in go 1.8.

horkhe:
If we rely on go 1.8 for HTTP/2 then it looks like it becomes a requirement to compile vulcand with 1.8 and therefore makes graceful logic pulled from manners redundant, don't you think?

archisgore:
Yup! That's the plan. I'm just staging changes for safety. I think there's some logic in Graceful Listener which is still necessary.

I looked up the go documentation and the default listener, even after a "Stop/Close" will continue to accept new connections (even in go 1.8)

And GracefulListener APIs are called by GracefulServer to off-load some processing.

Bringing it into vulcan means that I can at least make rapid changes and test rather than go through the change-godep-test cycle. I'm planning to remove whatever parts are redundant while keeping those that are not.

I'm currently in the process of setting up some functional tests on our production pipeline, which are intended to help me catch regressions.

horkhe:
@archisgore agree. Good job!

PR #337, opened 2017-05-31
Break proxy into subpackages

horkhe:
`vulcand/proxy` turned out to be a rather large and complex package. To make it easier to understand and deal with I split it into subpackages that helped leverage the compiler to enforce proper interface/implementation boundaries. It also got rid of cyclic dependencies between proxy components.


archisgore:
I like this! Thanks! Definitely will also help avoid naming conflicts.

PR #339, opened 2017-09-25
AutoCert (Let's Encrypt) Support.

archisgore:
This is the beginnings of enabling Let's Encrypt.


horkhe:
It would be nice if you move dependency update into a separate PR.

horkhe:
By the way what do you think about switching to [dep](https://github.com/golang/dep) for dependency management we already do that in Mailgun. It is still lives much to be desired, but at least it is supposed to become a standard tool 🤞.

archisgore:
Yes indeed, I'm going to break this up and do this step by step once it begins working. My plan is to first get vulcand/travis on go1.9, then update all dependencies required, and finally overlay this branch on top there. I also need to rebase quite a bit to make the change history readable and understandable.

About Dep - I just heard about it. I'm good with anything that's not Godep because it is seriously annoying! We have used glide in Polyverse for some time, and I'm down with dep. For better or for worse, if it's the new standard, let's stick to it.

horkhe:
@archisgore fyi, we at Mailgun are going to try routing gRPC requests with vulcand. In the next month I will be working on that. As the first step I will be adding unit tests to vulcand to see how gRPC requests behave.

horkhe:
@archisgore by the way `dep` might not work out-of-the box. I had experience dealing with it, so if you want I can take care of migrating to it.

archisgore:
I'll rebase off master once the godep branch is merged in. I merged this entire change in one commit. Here's the full description.

AutoCert Support (Let's Encrypt)
This change enables Let's Encrypt support through
use of the autocert package provided by golang.

Key things of Note:
1. A Host now has "AutoCert" as a field, in addition to KeyPair.
   When KeyPair is set, that key takes precedence to AutoCert.

   When AutoCert is set, it attempts an automatic Certificate
   generation for that hostname.

2. AutoCert field itself has a few members, namely:
   a) A DirectoryURL that allows you to set what ACME service to
      get certificates from (the ACME CA). Defaults to Let's Encrypt.

   b) A Key that is a PEM-encoded Private Key which authenticates
      you as a certificate-requestor to your ACME CA.
      A new 256-bit EC Key is generated if one is not provided.

   c) An Email that the CA can email you at, for failures or errors.

3. By default the Engines provide an AutoCertCache (where automatically
   generated certificates are stored.) This allows certificates to be
   cached along with Vulcan's own routing state.

   For instance, Etcd Engines will cache certificates in Etcd.

4. Caching is made a bit generic by providing an interface called
   a CacheProvider. This is intended for future use where other
   data might need to be cached, and it prevents a lot of re-wiring.

5. A different caching implementation may be set in the registry
   when initializing Vulcan (just like plugins), and this implementation,
   when non-nil, will be used in place of the Engine-based caches.

   This is very useful when testing vulcan itself and might require
   rapid restarts with the Memory Engine. Requesting the same
   certificate from Let's Encrypt repeatedly will easily
   get you banned.

   In this case, it might be worthwhile to use the DirCache implementation
   which caches on disk, instead of Etcd.

   This also enables potential caching to happen on remote/reliable
   KV stores such as Redis, DynamoDB, etc.

archisgore:
This is ready for code review. I'm going to update documentation first thing tomorrow.

archisgore:
This is ready to merge. I noticed Docs are in a different repo. I'll update those independently.

horkhe:
Got sick, will review in a couple of days, sorry.

horkhe:
@archisgore sorry for delay, it took me longer to recover then I expected. I am looking now. Have to read on ACME first to better understand what we are doing.

proxy/server/server.go
```diff
@@ -270,30 +281,91 @@ func (s *T) newTLSCfg(hostCfgs map[engine.HostKey]engine.Host) (*tls.Config, err
...
+
+			if ac.Key != "" || ac.DirectoryURL != "" {
+				autoCertMgr.Client = &acme.Client{
+					DirectoryURL: ac.DirectoryURL,
+				}
+
+				if ac.Key != "" {
+					block, _ := pem.Decode([]byte(ac.Key))
+					if block == nil {
+						log.Errorf("Autocert Key PEM Block for Host %s is invalid.", hostCfg.Name)
```
horkhe: Why do not return an error here and in the following checks?

proxy/server/server.go
```diff
@@ -270,30 +281,91 @@ func (s *T) newTLSCfg(hostCfgs map[engine.HostKey]engine.Host) (*tls.Config, err
...
+							log.Errorf("Error parsing Autocert Key block of type %s, for Host %s, as an EC Private Key: %v", block.Type, hostCfg.Name, err)
+						}
+					} else {
+						log.Errorf("AutoCert Private Key for Host %s is of unrecognized type: %s. Supported types"+
+							"are RSA PRIVATE KEY and EC PRIVATE KEY.", hostCfg.Name, block.Type)
+					}
+				}
+			}
+
+			cert, err := autoCertMgr.GetCertificate(&tls.ClientHelloInfo{ServerName: hostCfg.Name})
```
horkhe: How is it possible to get a certificate here if the server is not yet listening with this tls.Config? Who would respond to HTTP challenge?

service/service.go
```diff
@@ -402,6 +402,14 @@ func (s *Service) reportSystemMetrics() {
 }
 
 func (s *Service) newProxy(id int) (proxy.Proxy, error) {
+
+	cacheProvider := s.registry.GetCacheProvider()
+	// If there's no cache provider by the registry,
+	// then use one provided by the engine, if any.
+	if cacheProvider == nil {
```
horkhe: ... respectively we do not need this if you agree that couple of cache with engine is not necessary.

horkhe:
I completed the first pass. Sorry for delay and thanks a lot for doing this.

horkhe:
A general comment that in all files it would be nice to format imports with `goimport` so all imports are broken into two subsets sdk and the rest both alphabetically sorted.

horkhe:
Something to think about: at Mailgun we would like to be able to serve thousands of auto cert hosts on a distributed vulcand cluster. So it would be nice to be able to use an external `autocert.HostPolicy`.

plugin/cacheprovider/CacheProvider.go
```diff
@@ -0,0 +1,7 @@
+package cacheprovider
+
+import "golang.org/x/crypto/acme/autocert"
+
+type CacheProvider interface {
```
horkhe: Name it `T` so it can be used as `cacheprovider.T` to avoid statter?

engine/model.go
```diff
@@ -170,9 +170,17 @@ func (o *OCSPSettings) Equals(other *OCSPSettings) bool {
...
-	OCSP    OCSPSettings
+	Default  bool
+	KeyPair  *KeyPair
+	AutoCert *AutoCertSettings
+	OCSP     OCSPSettings
+}
+
+type AutoCertSettings struct {
+	Email              string
+	RenewBeforeSeconds int
```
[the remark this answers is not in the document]
archisgore: Ah yes of course. Makes more sense. I was just thinking that this setting will be done from etcd, but it makes more sense to just be consistent.

engine/engine.go
```diff
@@ -89,6 +90,12 @@ type Engine interface {
 	// GetRegistry returns registry with the supported plugins. It should be stored by Engine instance.
 	GetRegistry() *plugin.Registry
 
+	// The Engine can provide a CacheProvider, where it can offer to
+	// cache values in whatever storge the engine itself is backed by.
+	// This would allow especially etcd, but any future backend stores
+	// such as Consul, to cache values.
+	GetCacheProvider() cacheprovider.CacheProvider
```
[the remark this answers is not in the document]
archisgore: Okay! That works for me actually. Since in our implementation, I'd almost always only use an explicitly provided one. I'll do this too. Due to original discussion, I wanted to reuse engine, but that breaks several principles of proper decoupling. Thanks for the feedback!

proxy/mux/mux_test.go
```diff
@@ -370,6 +382,286 @@ func (s *ServerSuite) TestHostKeyPairUpdate(c *C) {
 	c.Assert(GETResponse(c, b.FrontendURL("/")), Equals, "Hi, I'm endpoint")
 }
 
+func (s *ServerSuite) TestServerHTTPSAutoCert(c *C) {
```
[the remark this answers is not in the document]
archisgore: Will do!

proxy/mux/mux_test.go
```diff
@@ -370,6 +382,286 @@ func (s *ServerSuite) TestHostKeyPairUpdate(c *C) {
...
+	e := testutils.NewHandler(func(w http.ResponseWriter, r *http.Request) {
+		req = r
+		w.Write([]byte("hi https"))
+	})
+	defer e.Close()
+
+	man := &autocert.Manager{
+		Prompt: autocert.AcceptTOS,
+	}
+	url, finish := startACMEServerStub(c, man, "example.org")
```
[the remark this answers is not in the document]
archisgore: I struggled with this one a lot. The problem with the staging environment is that you still need to be behind a real host with an internet-facing name to generate the cert. So it works for integration tests, but not Unit Tests. What I suggest is we keep the UTs in place, since they validate the basic logic, and then add integration tests as a followup to test real-world usage.

proxy/server/server.go
archisgore: I think the autocert manager within itself opens up a temporary endpoint. Somehow I've always found it able to get a cert each time I tried this.

service/service.go
archisgore: I'll just remove the engine coupling. :-) I didn't like it much anyway.

proxy/server/server.go
archisgore: Thanks for catching that. I'll just "continue" the loop at the failure.

archisgore:
About custom HostPolicy, I think we should do it organically. I'm not really clear on how to merge host policies from various places - are they additive? Exclusive? etc. etc. And Let's Encrypt will begin supporting wildcard certs from January 2018. So lots changing here soon. I just decided to keep the HostPolicy tightly bound to the specific host entered in config, that gives us the best base-level security posture for now.

proxy/server/server.go
archisgore: Actually none of my UTs really test for that. I'm removing it.

archisgore:
Okay following changes made:
1. Removed CacheProvider from engine entirely.
2. CacheProvider is currently only settable through Registry. This is to get core functionality in without adding additional CLI flags/wiring just yet. Let's us start testing.
3. By default MemCacheProvider is used, so certs won't be constantly regenerated on a single vulcan instance.
4. Rebased off master (hence the history rewriting and force-pushes.) This should make it FF-mergable onto master.
5. Also merged in the SNI UT fix, and removed the GetCertificate line that I was running at instantiation-time. GetCertificate will now only be called on the first request to that hostname.

proxy/server/server.go
```diff
@@ -270,30 +282,86 @@ func (s *T) newTLSCfg(hostCfgs map[engine.HostKey]engine.Host) (*tls.Config, err
...
+				HostPolicy: autocert.HostWhitelist(hostCfg.Name),
+				Email:      ac.Email,
 			}
+
+			if ac.RenewBeforeSeconds > 0 {
+				autoCertMgr.RenewBefore = time.Duration(ac.RenewBeforeSeconds) * time.Second
+			}
+
+			if ac.Key != "" || ac.DirectoryURL != "" {
+				autoCertMgr.Client = &acme.Client{
```
horkhe: If `ac.DirectoryURL != ""` and `ac.Key == ""` then we create a client but do not generate a key, and autoCertMgr won't do that for us. It only generates a key if `autoCertMgr.Client == nil`. We need a test that exposes this but, and then fix it.

proxy/server/server.go
```diff
@@ -7,6 +7,10 @@ import (
 	"net/http"
 	"sync"
```
horkhe: In a couple of files I noticed that stdlib section is split in two. We need to remove that empty line and re run `goimport` to sort them properly.

proxy/server/server.go
```diff
@@ -312,6 +380,22 @@ func (s *T) newTLSCfg(hostCfgs map[engine.HostKey]engine.Host) (*tls.Config, err
 	}
 
 	config.BuildNameToCertificate()
+
+	//Lock down this specific getCertFuncs map
```
horkhe: Here and in many other places `//` should be separated from the text with a space just to be consistent.

proxy/server/server.go
horkhe: Scratch that. After closely examining the AutoCert code I see that Client.Key is generate if it is empty in `autocert.(*autocert.Manager).acmeClient()`

proxy/server/server.go
```diff
@@ -270,30 +282,86 @@ func (s *T) newTLSCfg(hostCfgs map[engine.HostKey]engine.Host) (*tls.Config, err
...
+			keyPair, err := tls.X509KeyPair(c.Cert, c.Key)
 			if err != nil {
-				log.Warningf("%v failed to staple %v, error %v", s, hostCfg, err)
-			} else if r.Response.Status == ocsp.Good || r.Response.Status == ocsp.Revoked {
-				keyPair.OCSPStaple = r.Staple
-			} else {
-				log.Warningf("%s got undefined status from OCSP responder: %v", s, r.Response.Status)
+				return nil, err
+			}
+			if hostCfg.Settings.OCSP.Enabled {
```
horkhe: We only honor OCSP.Enabled for hosts with static certificates, I guess we can look into implementing that for AutoCert hosts in the scope of different PR.

proxy/mux/mux_test.go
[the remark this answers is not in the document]
horkhe: You are right, I did not thought

proxy/server/server.go
horkhe: How about changing this function a bit:
```go
func (s *T) newTLSCfg(hostCfgs map[engine.HostKey]engine.Host) (*tls.Config, error) {
	config, err := s.lsnCfg.TLSConfig()
	if err != nil {
		return nil, err
	}

	if config.NextProtos == nil {
		// "h2" is required in order to enable HTTP 2: https://golang.org/src/net/http/server.go
		config.NextProtos = []string{"h2", "http/1.1"}
	}

	defaultHostName := ""
	pairs := map[string]tls.Certificate{}
	getCertFuncs := map[string]getCertificateFunc{}

	for _, hostCfg := range hostCfgs {
		if hostCfg.Settings.Default {
			defaultHostName = hostCfg.Name
		}
		if hostCfg.Settings.KeyPair != nil {
			c := hostCfg.Settings.KeyPair
			if hostCfg.Settings.AutoCert != nil {
				log.Warningf("Host %v has a KeyPair and AutoCert enabled. KeyPair takes precedence. Autocert generation disabled.", hostCfg.Name)
			}
			keyPair, err := tls.X509KeyPair(c.Cert, c.Key)
			if err != nil {
				return nil, err
			}
			if hostCfg.Settings.OCSP.Enabled {
				log.Infof("%v OCSP is enabled for %v, resolvers: %v", s, hostCfg, hostCfg.Settings.OCSP.Responders)
				r, err := s.stapler.StapleHost(&hostCfg)
				if err != nil {
					log.Warningf("%v failed to staple %v, error %v", s, hostCfg, err)
				} else if r.Response.Status == ocsp.Good || r.Response.Status == ocsp.Revoked {
					keyPair.OCSPStaple = r.Staple
				} else {
					log.Warningf("%s got undefined status from OCSP responder: %v", s, r.Response.Status)
				}
			}
			pairs[hostCfg.Name] = keyPair
		} else if hostCfg.Settings.AutoCert != nil {
			ac := hostCfg.Settings.AutoCert
			if s.autoCertMgrs == nil {
				s.autoCertMgrs = make([]*autocert.Manager, 0, 0)
			}

			// Each host gets its own Autocert Manager - this allows individual
			// certs to use different autocert authorities, as well as auth keys
			autoCertMgr := &autocert.Manager{
				Client: &acme.Client{
					// Let's Encrypt will be used if ac.DirectoryURL is "".
					DirectoryURL: ac.DirectoryURL,
				},
				Prompt:     autocert.AcceptTOS,
				Cache:      s.autoCertCache,
				HostPolicy: autocert.HostWhitelist(hostCfg.Name),
				Email:      ac.Email,
			}

			if ac.RenewBeforeSeconds > 0 {
				autoCertMgr.RenewBefore = time.Duration(ac.RenewBeforeSeconds) * time.Second
			}

			if ac.Key != "" {
				block, _ := pem.Decode([]byte(ac.Key))
				if block == nil {
					log.Errorf("Autocert Key PEM Block for Host %s is invalid", hostCfg.Name)
					continue
				}
				switch block.Type {
				case "RSA PRIVATE KEY":
					autoCertMgr.Client.Key, err = x509.ParsePKCS1PrivateKey(block.Bytes)
					if err != nil {
						log.Errorf("Error parsing Autocert Key block of type %s, for Host %s, as an RSA Private Key: %v", block.Type, hostCfg.Name, err)
						continue
					}
				case "EC PRIVATE KEY":
					autoCertMgr.Client.Key, err = x509.ParseECPrivateKey(block.Bytes)
					if err != nil {
						log.Errorf("Error parsing Autocert Key block of type %s, for Host %s, as an EC Private Key: %v", block.Type, hostCfg.Name, err)
						continue
					}
				default:
					log.Errorf("AutoCert Private Key for Host %s is of unrecognized type: %s. Supported types"+
						"are RSA PRIVATE KEY and EC PRIVATE KEY.", hostCfg.Name, block.Type)
					continue
				}
			}

			getCertFuncs[hostCfg.Name] = autoCertMgr.GetCertificate
		}
	}

	// Add static certificates to the TLS config.
	config.Certificates = make([]tls.Certificate, 0, len(pairs))
	if defaultHostName != "" {
		keyPair, ok := pairs[defaultHostName]
		if !ok {
			return nil, errors.Errorf("default host '%s' certificate is not passed", defaultHostName)
		}
		config.Certificates = append(config.Certificates, keyPair)
	}
	for hostName, keyPair := range pairs {
		if hostName != defaultHostName {
			config.Certificates = append(config.Certificates, keyPair)
		}
	}
	config.BuildNameToCertificate()

	// Ensure generation/renewal of certificates for AutoCert enabled hosts.
	config.GetCertificate = func(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
		if getCertificateFunc, ok := getCertFuncs[info.ServerName]; ok {
			cert, err := getCertificateFunc(info)
			if err != nil {
				log.Errorf("Failed to generate Autocert for ServerName: %s. Error: %v.", info.ServerName, err)
			}
			return cert, err
		}
		return nil, nil
	}

	return config, nil
}
```

horkhe:
Overall looks very good, just minor formatting comments (sorry for being too pedantic) and a proposal to change the `newTLSCfg` to make it a bit shorter and easier to read. Almost ready to merge.

archisgore:
I broke it up into smaller functions. Much easier to read now. Also was able to get OCSP stapling done for AutoCert.

proxy/server/server.go
```diff
@@ -357,3 +378,133 @@ func newScopeRouter(scope string, router http.Handler) (http.Handler, error) {
...
+		// we support. Go's crypto library doesn't support a generic function to provide back
+		// a private key interface.
+		if ac.Key != "" {
+			block, _ := pem.Decode([]byte(ac.Key))
+			if block == nil {
+				return nil, fmt.Errorf("Autocert Key PEM Block for Host %s is invalid.", hostCfg.Name)
+			} else if block.Type == "RSA PRIVATE KEY" {
+				rsaPrivateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+				if err != nil {
+					autoCertMgr.Client.Key = rsaPrivateKey
```
horkhe: if error then assign?

proxy/server/server.go
```diff
@@ -357,3 +378,133 @@ func newScopeRouter(scope string, router http.Handler) (http.Handler, error) {
...
+				} else {
+					log.Errorf("Error parsing Autocert Key block of type %s, for Host %s, as an RSA Private Key: %v", block.Type, hostCfg.Name, err)
+				}
+			} else if block.Type == "EC PRIVATE KEY" {
+				ecPrivateKey, err := x509.ParseECPrivateKey(block.Bytes)
+				if err != nil {
+					autoCertMgr.Client.Key = ecPrivateKey
```
horkhe: if error then assign?

archisgore: Haha! OMG! I don't know how I missed that. Actually let me add a unit test for an EC key, along with this fix. Currently adding an OCSP Autocert UT as well.

archisgore:
I had to modify some stapler code to allow it to hold a GetCertificate function for the host, in addition to a KeyPair. This is in case the cert expires and GetCertificate changes its response - stapler must obtain a fresh OCSP response in that case for the new keypair.

This should be significantly mitigated through caching though.

proxy/server/server.go
archisgore: Added two UTs to guard against both code paths - RSA and EC Private Keys.

proxy/server/server.go
```diff
@@ -357,3 +377,144 @@ func newScopeRouter(scope string, router http.Handler) (http.Handler, error) {
...
+			DirectoryURL: ac.DirectoryURL,
+		}
+
+		// If Key is non-empty, then decode it as RSA or EC which are the only two keys
+		// we support. Go's crypto library doesn't support a generic function to provide back
+		// a private key interface.
+		if ac.Key != "" {
+			block, _ := pem.Decode([]byte(ac.Key))
+			if block == nil {
+				return nil, fmt.Errorf("Autocert Key PEM Block for Host %s is invalid.", hostCfg.Name)
```
horkhe: Could you please use `github.com/pkg/errors.New/Errorf/Wrap(f)` whenever you return an error. That will ensure that stack information is available in error objects to be logged later, and in case of `errors.Wrapf` the chain of errors is explicit.

proxy/server/server.go
```diff
@@ -357,3 +377,144 @@ func newScopeRouter(scope string, router http.Handler) (http.Handler, error) {
...
+			} else if block.Type == "RSA PRIVATE KEY" {
+				rsaPrivateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+				if err == nil {
+					autoCertMgr.Client.Key = rsaPrivateKey
```
horkhe: Why did you chose to ignore this error? As I understand a new key will be generated in this case. If it is indeed what you expect then please add a comment explaining that ignoring the error is deliberate choice.

This also goes for the EC Private Key.

proxy/server/server.go
```diff
@@ -357,3 +377,144 @@ func newScopeRouter(scope string, router http.Handler) (http.Handler, error) {
...
+
+	for hostName, keyPair := range pairs {
+		if hostName != defaultHostName {
+			arr = append(arr, keyPair)
+		}
+	}
+	return arr, nil
+}
+
+func ocspStapleToCert(stapler stapler.Stapler, hostCfg engine.Host, keyPair *tls.Certificate) {
```
horkhe: Code with nested if and if else is hard to read. It is more Golang idiomatic to return as soon as you have a chance, How about:
```Go
func ocspStapleToCert(stapler stapler.Stapler, hostCfg engine.Host, keyPair *tls.Certificate) {
	if !hostCfg.Settings.OCSP.Enabled {
		return
	}
	log.Infof("OCSP is enabled for %v, resolvers: %v", hostCfg, hostCfg.Settings.OCSP.Responders)

	r, err := stapler.StapleHost(&hostCfg)
	if err != nil {
		log.Warningf("Failed to staple %v, error %v", hostCfg, err)
		return
	} 
	if r.Response.Status != ocsp.Good && r.Response.Status != ocsp.Revoked {
		log.Warningf("Got undefined status from OCSP responder: %v", r.Response.Status)
		return
	}
	keyPair.OCSPStaple = r.Staple
}
```

proxy/server/server.go
```diff
@@ -357,3 +377,144 @@ func newScopeRouter(scope string, router http.Handler) (http.Handler, error) {
...
+			}
+		}
+	}
+
+	// Wrap the GetCert for this host, so we can generate and staple
+	// an optional OCSP response when requested.
+	stapledGetCert := func(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
+		keyPair, err := autoCertMgr.GetCertificate(info)
+
+		if hostCfg.Settings.OCSP.Enabled {
```
horkhe: Can't we just call `ocspStapleToCert` here? The code seems identical.

horkhe:
Almost there :)

proxy/server/server.go
archisgore: Ah! I didn't even know about this! Thanks for telling me.

proxy/server/server.go
archisgore: Oops. This came from previous logic where this was in a for-loop. I need to return both.

proxy/server/server.go
archisgore: Did that...

proxy/server/server.go
archisgore: Aha. I am fixing that. The reason for passing the GetCertFunc into stapler is when the stapler updates the OCSP payload, the cert might have changed. I added an extra param to ocspStapleToCert to pass that function through.

archisgore:
Ready for next review. Thanks for pointing out I was ignoring the error. The UTs were passing despite a failure in fact. I had to update the EC key once the code was fixed.

horkhe:
Thank you!

PR #343, opened 2017-10-11
Switch to dep for vendoring

horkhe:
Unfortunately [dep](https://github.com/golang/dep) is yet far from perfection. Even though it properly deduces all transitive dependencies `dep prune` does not clean some unused code, that may cause compilation errors. To workaround this deficiency `scripts/vendor.sh` script is provided that is supposed to be used instead of `dep ensure`.

It seems like a lot of has been changed but the lion share of new files are non-code files that used to be purged by godep.


horkhe:
@archisgore that was fast :) thanks

PR #344, opened 2017-10-11
Fix SNI Test to work with multiple hosts correctly

archisgore:
When testing AutoCert, I realized that this test was not comprehensive enough.

If we add an additional host call, say "foobar", we get a 404 back because the
route doesn't exist for it. However, the default cert is returned regardless.

This means that if we checked for the Certificate Serial Number on the returned
cert, it is constant across any host name. In short, exposing multiple SSL
Certs by Host, will not result in different certs being returned
based on the request made.

I have set up the initial conditions for the test which fails right now.

I'll figure out how to make it pass.

### DIFF
--- a/proxy/mux/mux_test.go
+++ b/proxy/mux/mux_test.go
@@ -457,7 +457,7 @@ func (s *ServerSuite) TestSNI(c *C) {
 	b := MakeBatch(Batch{
 		Host:     "localhost",
 		Addr:     "localhost:41000",
-		Route:    `Host("localhost") && Path("/")`,
+		Route:    `Path("/path1")`,
 		URL:      e.URL,
 		Protocol: engine.HTTPS,
 		KeyPair:  &engine.KeyPair{Key: localhostKey, Cert: localhostCert},
@@ -465,7 +465,7 @@ func (s *ServerSuite) TestSNI(c *C) {
 	b2 := MakeBatch(Batch{
 		Host:     "otherhost",
 		Addr:     "localhost:41000",
-		Route:    `Host("otherhost") && Path("/")`,
+		Route:    `Path("/path2")`,
 		URL:      e2.URL,
 		Protocol: engine.HTTPS,
 		KeyPair:  &engine.KeyPair{Key: otherHostKey, Cert: otherHostCert},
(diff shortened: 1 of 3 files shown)


proxy/mux/mux_test.go
```diff
@@ -452,26 +457,33 @@ func (s *ServerSuite) TestSNI(c *C) {
...
+		KeyPair:  &engine.KeyPair{Key: otherHostKey, Cert: otherHostCert},
 	})
 	b2.H.Settings.Default = true
 	c.Assert(s.mux.Init(MakeSnapshot(b, b2)), IsNil)
 	c.Assert(s.mux.Start(), IsNil)
 
-	// When/Then
-	c.Assert(GETResponse(c, b.FrontendURL("/"), testutils.Host("localhost")), Equals, "Hi, I'm endpoint 1")
-	c.Assert(GETResponse(c, b.FrontendURL("/"), testutils.Host("otherhost")), Equals, "Hi, I'm endpoint 2")
+	// For the same host, but different paths, return different endpoints (routing test)
```
horkhe: No need to test routing here, it is tested in other tests, and just clutters this one.

proxy/mux/mux_test.go
```diff
@@ -452,26 +457,33 @@ func (s *ServerSuite) TestSNI(c *C) {
...
+	c.Assert(GETResponse(c, b.FrontendURL("/path1"), testutils.Host("example.com")), Equals, "Hi, I'm endpoint 1")
+	c.Assert(GETResponse(c, b.FrontendURL("/path2"), testutils.Host("example.com")), Equals, "Hi, I'm endpoint 2")
+
+	//For the same path, if the Hostname is different (SNI), then return a different Cert - true host differentiation.
+	c.Assert(GETPeerCertSerialNo(c, b.FrontendURL("/path1"), testutils.Host("example.com")), Equals, "77bdc3e97d00584f03faec7cda682cf")
+	c.Assert(GETPeerCertSerialNo(c, b.FrontendURL("/path1"), testutils.Host("otherhost")), Equals, "c3244866e57c7b1f")
```
horkhe: Should not this fail because `path1` is not defined for otherhost?

proxy/mux/mux_test.go
```diff
@@ -1136,6 +1148,15 @@ func GETResponse(c *C, url string, opts ...testutils.ReqOption) string {
 	return string(body)
 }
 
+func GETPeerCertSerialNo(c *C, url string, opts ...testutils.ReqOption) string {
```
horkhe: It is only used locally so the name should not be capitalized `getPeerCertSerialNo`.

proxy/mux/mux_test.go
horkhe: I guess the only thing that shit test should verify is that we serve correct TLS certificate based on the host. Host based routing is checked in another test. So for that we do not even need two responders. Because all we care is the cert no.

archisgore:
Done! Thanks for pointing that out. Makes tests cleaner.

PR #347, opened 2017-11-16
Add option to use memng engine

nyanshak:
Allows users to specify configuration engine, either memng or etcd. Defaults to etcd to maintain backwards-compatible behavior. Specified with `-engine $engine`


nyanshak:
@archisgore wow, thanks! Didn't expect a response so quickly :1st_place_medal: 

I tried running the tests locally (both with 1.9.2 and 1.9 (downgraded my version in case that was the issue)) and I'm seeing the same failures on master as I am on my branch. I think there might be some problem with the tests themselves; could you take a look? I don't think I've changed anything that would have broken these if they were working to begin with, but I'm not sure.

Issue #348, opened 2017-12-06
update route path error


onestraw:
# config
```bash

export ETCDCTL_API=2
etcdctl set /vulcand/backends/b1/backend '{"Type": "http"}'
etcdctl set /vulcand/backends/b1/servers/srv1 '{"URL": "http://localhost:8001"}'
etcdctl set /vulcand/backends/b1/servers/srv2 '{"URL": "http://localhost:8002"}'
etcdctl set /vulcand/frontends/f1/frontend '{"Type": "http", "BackendId": "b1", "Route": "Path(`/`)"}'

# start simple server at port 8000
python -m SimpleHTTPServer 8001
python -m SimpleHTTPServer 8002

# access the frontend service localhost:8181
curl localhost:8181

```
# bug

```bash
etcdctl set /vulcand/frontends/f1/frontend '{"Type": "http", "BackendId": "b1", "Route": "Path(`/123`)"}'
etcdctl set /vulcand/frontends/f1/frontend '{"Type": "http", "BackendId": "b1", "Route": "Path(`/`)"}'

curl localhost:8181
{"error":"not found"}%
```

# fix
proxy/frontend/frontend.go: Update()

```
- if !feCfg.HTTPSettings().Equals(fe.cfg.HTTPSettings()) {
+ if !feCfg.HTTPSettings().Equals(fe.cfg.HTTPSettings()) || fe.cfg.Route != feCfg.Route {
```

horkhe:
Thank you. If you could turn it into a PR that would be great.

onestraw:
I have sent the PR, pls review

PR #349, opened 2017-12-07
fix the route path update error

onestraw:
- fix issue #348
       - add the Equals method for Frontend

Issue: #348


horkhe:
Thank you. The only thing that is missing is a [test] (https://github.com/vulcand/vulcand/blob/master/proxy/mux/mux_test.go) that fails without this change and passes with it.

horkhe:
Thank you

PR #350, opened 2018-02-22
Added -localhost-alias option

thrawn01:
## Purpose
When running vulcand in a kubernetes `DaemonSet` vulcand needs to know requests from the local node should match `Host("localhost")` rules. This flag allows an author of a vulcand `DaemonSet` to tell vulcand the name of the node it's currently running on, such that vulcand correctly routes requests for `Host("localhost")` 

## Implementation
* Simple search and replace for `Route` rules and add additional route handler with the alias host.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 flymake_*
 *.swp
 *.go~
+.idea/
--- a/proxy/mux/mux.go
+++ b/proxy/mux/mux.go
@@ -2,6 +2,7 @@ package mux
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"sort"
 	"strconv"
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 flymake_*
 *.swp
 *.go~
+.idea/
```
archisgore: I think you can remove the IDE metadata. Add to .gitignore perhaps.

thrawn01:
I started down the path of modifying https://github.com/vulcand/route first, my original thought was I would make a router that, if it saw a predicate `Host("<some-value>")` it would return a matcher equivalent to `Host("<some-value>") || Host("<alias>")` however when I tried this, I noted `||` isn't supported for trie hosts. So I don't see how I can safely modify the router to return an appropriate matcher for this scenario. What I had to do was add an *additional* matcher for the alias.

.gitignore
thrawn01: ??

archisgore: Haha. Sorry. :-) My bad.

archisgore:
What I'm proposing is, if 
```
func (m *mux) AddFrontendHandler(expr string, handler http.Handler) error {
```

Only happened inside 

```
router.Handle(expr, handler)
```

So the router could still replace the expr in-place before it stored it:

https://github.com/vulcand/route/blob/master/mux.go#L24

This
```
func (m *Mux) Handle(expr string, handler http.Handler) error {
	return m.router.UpsertRoute(expr, handler)
}
```

could be:
func (m *Mux) Handle(expr string, handler http.Handler) error {
        // if m.somevalue -> rewrite expr -> proceed to UpsertRoute
	return m.router.UpsertRoute(expr, handler)
}

thrawn01:
This experiment was abandoned, we are moving in a new direction =)

PR #354, opened 2018-07-04
Add support for TLS and Auth for ETCD engines

thrawn01:
## Purpose
Mailgun is migrating our infrastructure to ETCD v3 with TLS and Auth. Before this migration occurs Vulcand needs to support TLS and Auth for the ETCD v3 engine. This change introduces auth and basic TLS for both v3 and v2 etcd engines.

## Implemention
* Moved `etcdv3ng` and `etcdv2ng` packages to `etcdng/v2` and `etcdng/v3` this allows both engines to share common code in `etcdng`
* Removed duplicate `Options` structs in both versions and renamed to `etcdng.Options`. Both v3 and v2 now use the same struct for their options.
* Added `etcdng.NewTLSConfig()` and is now used by both v3 and v3 engines
* Added the following command line arguments `-etcdUsername, -etcdPassword -etcdInsecureSkipVerify -etcdEnableTLS -etcdDebug`
* Introduced environment variables `VULCAND_TEST_ETCD_USER, VULCAND_TEST_ETCD_PASS, VULCAND_TEST_ETCD_DEBUG, VULCAND_TEST_ETCD_TLS` for auth and TLS setup during tests.
* Subscribe not longer reads the host entry from etcd instead of using the value provided in the v3 event. This would cause random test failures for the  `TestHostUpsertKeyPair` test


service/service.go
```diff
@@ -354,32 +355,32 @@ func (s *Service) newEngine() error {
...
+		Password:            s.options.EtcdPassword,
+		InsecureSkipVerify:  s.options.EtcdInsecureSkipVerify,
+		EnableTLS:           s.options.EtcdEnableTLS,
+		Debug:               s.options.EtcdDebug,
+		Box:                 box,
+	}
+
 	if s.options.EtcdApiVersion == 3 {
-		ng, err = etcdv3ng.New(
+		ng, err = v3.New(
```
horkhe: In this context `v3` is not descriptive enough. I think we better use import aliases (e.g. etcd3ng) or something like that.

engine/etcdng/v3/etcd.go
```diff
@@ -14,10 +14,13 @@ import (
 	"github.com/coreos/etcd/mvcc/mvccpb"
 	log "github.com/sirupsen/logrus"
 	"github.com/vulcand/vulcand/engine"
+	"github.com/vulcand/vulcand/engine/etcdng"
 	"github.com/vulcand/vulcand/plugin"
 	"github.com/vulcand/vulcand/secret"
 	"github.com/vulcand/vulcand/utils/json"
 	"golang.org/x/net/context"
+	"google.golang.org/grpc/grpclog"
+	"os"
```
horkhe: Could you please format files with `goimport`.

horkhe:
> Moved etcdv3ng and etcdv2ng packages to etcdng/v2 and etcdng/v3 this allows both engines to share common code in etcdng

This is great. I wanted to do that long time ago. Although I think way more code belongs to the generic `etcdng`. All data parsing functions (e.g. GetSnapshot, parseBackend, parseFrontend, etc) I think can go there.

thrawn01:
@horkhe Looking at the methods you point to in both packages they are different enough that it doesn't make sense to move them. However, there are some private methods that are identical that call un-identical private methods. The composition features of golang wouldn't allow me to move them into a separate struct and then compose them such that the composed struct would call the encapsulating private methods. I feel this is a plus because polymorphism can be a pain to follow. As such, I feel good in leaving those methods altho identical in separate packages. There is minimal overlap that I can see. Let me know if you see some methods I missed.

PR #355, opened 2018-07-07
Revendored oxy to latest version

archisgore:
As part of this, I had to set some dependencies to:
```
branch = "master"
```

to ensure that oxy's dependencies flowed down into vulcan.

One minor code change converted `pw.Code` to `pw.StatusCode()`.

Otherwise just a `dep ensure -update` did the trick. Builds and tests pass.


archisgore:
Damn. I'll fix the UTs.

archisgore:
Ready to merge.

Gopkg.toml
```diff
@@ -50,7 +50,7 @@
   name = "github.com/mailgun/metrics"
 
 [[constraint]]
-  revision = "fd192d755b00c968d312d23f521eb0cdc6f66bd0"
+  branch = "master"
```
horkhe: This is basically means that there is no constraint, so maybe it is better to remove constraints with `branch="master"` all together?

archisgore: Yup. Will do. Good point.

archisgore:
@horkhe Done. I removed other "master" branch constraints while at it. Revendored all files.

I kept the gopkg.toml updates in a separate commit from all the vendored changes. So it should be easy to review.w

PR #360, opened 2018-09-12
Added timeout while establishing etcd v3 watch

thrawn01:
## Purpose
While using the etcd v3 protocol watches can silently timeout or fail to establish a watch. 

* https://github.com/etcd-io/etcd/pull/10019 
* https://github.com/etcd-io/etcd/pull/10020

## Implementation
In the case etcd v3 `Watch()` blocks indefinitely vulcand should error and try again. As Vulcand already retries when `init()` returns an error.  `Subscribe()` now returns an error when `watcher.Watch()` takes to long to return.

Also added `WithRequireLeader()` to avoid etcd network splits from blocking watches indefinitely (See https://github.com/etcd-io/etcd/issues/7593#issuecomment-288930039)

### DIFF
--- a/engine/etcdng/v3/etcd.go
+++ b/engine/etcdng/v3/etcd.go
@@ -597,9 +597,19 @@ func (n *ng) Subscribe(changes chan interface{}, afterIdx uint64, cancelC chan s
 	watcher := etcd.NewWatcher(n.client)
 	defer watcher.Close()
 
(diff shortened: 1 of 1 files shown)


archisgore:
This is funny and good catch! I didn't even know this was a thing. I assumed Watch was a quick/async method and if it failed, the watchChan would be closed. You learn something new every day. I love the change. Thank you.

PR #361, opened 2018-09-12
etcdv3 now correctly loads middleware

thrawn01:
## Purpose
Middlewares are not loaded when using the etcd v3 engine.

## Implementation
Currently, the code filters retrieved keys by `/vulcand/frontends/<route>/frontend/middlewares` which is incorrect. This PR filters the retrieved keys by `/vulcand/frontends/<route>/middlewares`


archisgore:
LGTM and I'm approving this. Would love it if you could add a small UT to guard this. Clearly I missed this in my implementation. It would help the next person (if say Consul or some other backend was ever added) to not miss this use-case.

PR #370, opened 2019-10-31
Add memng engine option

nyanshak:
* Allow users to optionally specify the engine with `-engine $engine`.
Defaults to etcd engine to maintain current default behavior, but allows
`memng` option.


nyanshak:
Re-opening from https://github.com/vulcand/vulcand/pull/347, re-created off master since it hadn't been updated in so long and so much had changed.

PR #372, opened 2020-08-14
SRE-542: Return error when watcher channel closes unexpectedly

pirogoeth:


thrawn01:
We should also return a non nil on line 619, The documentation for `Cancelled` says this.
```
// Canceled is used to indicate watch failure.
// If the watch failed and the stream was about to close, before the channel is closed,
// the channel sends a final response that has Canceled set to true with a non-nil Err().
```
So it's possible etcd was signaling the watch was cancelled but we didn't retry to restart the watch.

engine/etcdng/v3/etcd.go
```diff
@@ -615,12 +615,14 @@ func (n *ng) Subscribe(changes chan interface{}, afterIdx uint64, cancelC chan s
...
 		if response.Canceled {
-			log.Infof("stop watching: graceful shutdown")
-			return nil
-		}
-		if err := response.Err(); err != nil {
-			log.Errorf("stop watching: error: %v", err)
-			return err
+			log.Warn("etcd watcher cancelled")
+
+			if err := response.Err(); err != nil {
```
thrawn01, line 620: In practice I've found that we can get an error without `Cancelled = true`. So we should check for it separately, but log it if it's set when `Cancelled = true`. (etcd watchers are in such a horrible state)

pirogoeth: ok :+1:

PR #378, opened 2022-10-06
Update vulcand dependencies

thrawn01:
### Purpose
Resolve some security issues found in dependent packages and upgrade the etcd client to the latest version to take advantage of some improvements in how watches handle errors. https://github.com/etcd-io/etcd/pull/13999

### Implementation
* Lint improvements to etcdv3 implementation
* Removed tests for `autocert` as the tests only understand pre https://datatracker.ietf.org/doc/html/rfc8555 acme server. As the standard for acme has changed completely the current implementation vulcand uses is broken. Since this been broken for some time, it is safe to assume this feature is not used by anyone. We might consider removing this feature in a future release.
* Fixed `X-Forward-Host` header test which incorrectly tested for the backend host value instead of the client provided host value.
* Vendored an old version of `github.com/vulcand/oxy/ratelimit` into `plugin/ratelimit` as oxy has moved on to a completely different version, and updating to the latest oxy looks like it would require a complete redesign of the ratelimits package.
* Removed need for `metrics.NewNop()`which may have caused memory leak when not using metrics
* Fixed the `systests`
* Enabled the etcd3 tests during CI
* Etcd engine now defaults to v3 instead of v2


Takumi2008 (approved):

alanmmock (approved):

Baliedge (approved):

Takumi2008 (approved):

alanmmock (approved):

thrawn01:
Closing in favor of https://github.com/vulcand/vulcand/pull/379

PR #380, opened 2022-10-26
Improve how vulcand handles some errors in config

thrawn01:
### Purpose
To improve how vulcand reacts to various error conditions in the config. Depends on #379 

### Implementation
* Vulcand now returns 503 instead of 500 when no backend servers are defined. In addition it will log this error so users can alert on this message.


proxy/mux/mux.go
```diff
@@ -766,6 +774,159 @@ func (m *mux) TopServers(beKey *engine.BackendKey) ([]engine.Server, error) {
...
+
+// HealthCheckServers will run health checks on all servers, this method blocks until it
+// finds all servers for a backend are unhealthy for a period of time or until `done` is closed.
+//
+//	NOTE: This health check does NOT mark individual servers as disabled or remove them from
+//	the backend server list if they fail a health check. It is up to the backend config
+//	registrations to remove backend servers in a timely manner. This check is purely to warn
+//	operators that backends have failed and to reload the config in case no backend servers
+//	are healthy.
+func (m *mux) HealthCheckServers(done chan struct{}, opts proxy.HealthCheckOptions) error {
```
MatthewEdge, line 810: Reviewer note: you may notice that this method never actually returns an error for https://github.com/vulcand/vulcand/pull/380/files#diff-76859eda6768ff4ed42d3626ca7485696b179643131278e82ba295f5a9240bbeR262 to handle. This is intentional - we only log at this time while we determine how we want to handle the invalid backends

johngodsey (approved):

PR #384, opened 2024-07-27
Test for supported Go versions

HaraldNordgren:
Test for supported Go versions and utilizing the new “stable” and “oldstable” version aliases to avoid having to update this again for each successive Go version.

### DIFF
--- a/.github/workflows/on_pull_request.yaml
+++ b/.github/workflows/on_pull_request.yaml
@@ -15,7 +15,7 @@ jobs:
     name: test
     strategy:
       matrix:
-        go-version: [ 1.18.x, 1.19.x ]
+        go-version: [ oldstable, stable ]
         os: [ ubuntu-latest ]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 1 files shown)


HaraldNordgren:
Ping @MatthewEdge

.github/workflows/on_pull_request.yaml
```diff
@@ -15,15 +15,15 @@ jobs:
     name: test
     strategy:
       matrix:
-        go-version: [ 1.18.x, 1.19.x ]
+        go-version: [ oldstable, stable ]
```
MatthewEdge, line 18: Apologies for the delay. VulcanD isn't supported much these days and I didn't get the ping.

I'm not sure this is actually helpful for a project like VulcanD given how little activity it gets. The versions will likely skip many times before a new PR comes along to actually test new versions. If it broke somewhere in-between, how will you know?

HaraldNordgren: Fair point. Let's close it then 🤗

